PR #4, opened 2023-10-18
migrate Prover functionality from core repo

ChaoticWalrus:
had to modify some paths / folder structure / etc

Marking this PR as a draft for the moment, as it's currently having compiler issues due to duplicate instances of `IPauserRegistry.sol`.


stevennevins (approved):

ChaoticWalrus:
So this is now running as well as it was in the other repo.
The existing rules are failing, but due to 'havoc' behavior which really shouldn't be present, given the dispatching behavior defined in the spec.
We can merge this but it'll result in a failing CI for the moment, so... I'm fine either way 🤷

ChaoticWalrus:
closing this PR as stale; if these rules are added, it should be in a new PR

PR #21, opened 2023-10-24
Refactor StakeRegistry to use a signed delta for stake updates

wadealexc:
This PR:
* changes for loop index names to things like "i" or "j" when the index is just an index. The wordy index names were bugging me.
* Refactors StakeRegsitry to use a signed delta in most places, which allows us to get rid of a lot of bloat

Functional changes in StakeRegistry:
* Total stake updates are only made if we have a nonzero stake delta. Otherwise we skip them.
* For both total and operator updates, if the most recent entry was made during this block, we update that entry rather than pushing a new one. This should mean that all histories only have one entry per block.

### DIFF
--- a/src/BLSRegistryCoordinatorWithIndices.sol
+++ b/src/BLSRegistryCoordinatorWithIndices.sol
@@ -449,7 +449,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
             socket: socket
         });
         
-        for (uint i = 0; i < numOperatorsPerQuorum.length; i++) {
+        for (uint256 i = 0; i < numOperatorsPerQuorum.length; i++) {
             require(
                 numOperatorsPerQuorum[i] <= _quorumOperatorSetParams[uint8(quorumNumbers[i])].maxOperatorCount,
                 "BLSRegistryCoordinatorWithIndices._registerOperatorWithCoordinatorAndNoOverfilledQuorums: quorum is overfilled"
@@ -586,8 +586,8 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     ) external view returns (uint32[] memory) {
         uint32[] memory indices = new uint32[](operatorIds.length);
         for (uint256 i = 0; i < operatorIds.length; i++) {
-            uint length = _operatorIdToQuorumBitmapHistory[operatorIds[i]].length;
-            for (uint j = 0; j < length; j++) {
+            uint256 length = _operatorIdToQuorumBitmapHistory[operatorIds[i]].length;
+            for (uint256 j = 0; j < length; j++) {
                 if (_operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].updateBlockNumber <= blockNumber) {
                     uint32 nextUpdateBlockNumber = 
                         _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber;
--- a/src/BLSRegistryCoordinatorWithIndices.sol
+++ b/src/BLSRegistryCoordinatorWithIndices.sol
@@ -586,15 +586,16 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     ) external view returns (uint32[] memory) {
         uint32[] memory indices = new uint32[](operatorIds.length);
         for (uint256 i = 0; i < operatorIds.length; i++) {
-            uint32 length = uint32(_operatorIdToQuorumBitmapHistory[operatorIds[i]].length);
-            for (uint32 j = 0; j < length; j++) {
+            uint length = _operatorIdToQuorumBitmapHistory[operatorIds[i]].length;
+            for (uint j = 0; j < length; j++) {
                 if (_operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].updateBlockNumber <= blockNumber) {
+                    uint32 nextUpdateBlockNumber = 
+                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber;
                     require(
-                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber == 0 ||
-                        _operatorIdToQuorumBitmapHistory[operatorIds[i]][length - j - 1].nextUpdateBlockNumber > blockNumber,
+                        nextUpdateBlockNumber == 0 || nextUpdateBlockNumber > blockNumber,
                         "BLSRegistryCoordinatorWithIndices.getQuorumBitmapIndicesByOperatorIdsAtBlockNumber: operatorId has no quorumBitmaps at blockNumber"
                     );
-                    indices[i] = length - j - 1;
+                    indices[i] = uint32(length - j - 1);
                     break;
                 }
             }
--- a/src/StakeRegistry.sol
+++ b/src/StakeRegistry.sol
@@ -103,13 +103,8 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
                 }
             }
 
-            // If we have a change in total stake for this quorum, update state
-            // TODO - do we want to record the update, even if the delta is zero?
-            //        maybe this makes sense in the case that the quorum doesn't have
-            //        an update for this block?
-            if (totalStakeDelta != 0) {
-                _recordTotalStakeUpdate(quorumNumber, totalStakeDelta);
-            }
+            // Record the update for the quorum's total stake
+            _recordTotalStakeUpdate(quorumNumber, totalStakeDelta);
 
             unchecked {
                 ++quorumNumber;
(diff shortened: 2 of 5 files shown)


src/StakeRegistry.sol
```diff
@@ -154,39 +143,31 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
-            (, uint96 stake) = _updateOperatorStake({
+
+        for (uint i = 0; i < quorumNumbers.length; ) {            
+            /**
+             * Update the operator's stake for the quorum and retrieve their current stake
+             * as well as the change in stake.
+             * If this method returns `stake == 0`, the operator has not met the minimum requirement
+             * 
+             * TODO - we only use the `stake` return here. It's probably better to use a bool instead
+             *        of relying on the method returning "0" in only this one case.
```
wadealexc, line 154: Question for reviewers - see comment. I think it'd be a lot cleaner to switch the second return to a `bool` and use that here instead.

I think using a bool to be explicit is much better than treating "0" as an error value.

src/StakeRegistry.sol
```diff
@@ -283,71 +251,111 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
-        operatorIdToStakeHistory[operatorId][quorumNumber].push(operatorStakeUpdate);
-        return stakeBeforeUpdate;
+        
+        // Create a new stake update and push it to storage
+        // TODO - update the entry instead of pushing, if the last update block is this block
+        operatorIdToStakeHistory[operatorId][quorumNumber].push(OperatorStakeUpdate({
+            updateBlockNumber: uint32(block.number),
+            nextUpdateBlockNumber: 0,
+            stake: newStake
+        }));
```
wadealexc, line 312: Question for reviewers - currently, our history gets pushed to every time we do an update, even if the last update was the same block. 

This can lead to chains in the history where multiple history updates that occur in the same block have a "nextUpdateBlockNum" field that points to the same block as the next update.

I feel like it makes more sense to just check if the last update is the current block, and update that entry's stake if so.

test/harnesses/StakeRegistryHarness.sol
```diff
@@ -49,35 +49,31 @@ contract StakeRegistryHarness is StakeRegistry {
...
+
+        for (uint i = 0; i < quorumNumbers.length; ) {            
+            /**
+             * Update the operator's stake for the quorum and retrieve their current stake
+             * as well as the change in stake.
+             * If this method returns `stake == 0`, the operator has not met the minimum requirement
+             * 
+             * TODO - we only use the `stake` return here. It's probably better to use a bool instead
+             *        of relying on the method returning "0" in only this one case.
+             */
```
wadealexc, line 61: Question for reviewers - currently, our history gets pushed to every time we do an update, even if the last update was the same block. 

This can lead to chains in the history where multiple history updates that occur in the same block have a "nextUpdateBlockNum" field that points to the same block as the next update.

I feel like it makes more sense to just check if the last update is the current block, and update that entry's stake if so.

src/StakeRegistry.sol
wadealexc, line 312: Question for reviewers - currently, our history gets pushed to every time we do an update, even if the last update was the same block. 

This can lead to chains in the history where multiple history updates that occur in the same block have a "nextUpdateBlockNum" field that points to the same block as the next update.

I feel like it makes more sense to just check if the last update is the current block, and update that entry's stake if so.

src/StakeRegistry.sol
```diff
@@ -79,53 +79,42 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
-            if (totalStakeUpdate.updateBlockNumber != 0) {
-                // update the total stake history for the quorum
-                _recordTotalStakeUpdate(quorumNumber, totalStakeUpdate);
+
+            // If we have a change in total stake for this quorum, update state
+            // TODO - do we want to record the update, even if the delta is zero?
+            //        maybe this makes sense in the case that the quorum doesn't have
+            //        an update for this block?
+            if (totalStakeDelta != 0) {
+                _recordTotalStakeUpdate(quorumNumber, totalStakeDelta);
```
wadealexc, line 111: Question for reviewers - this one is similar to the other Q highlighted in this PR. Currently, the semantics here are a little weird - we basically push a stake update for a quorum if an operator belongs to the quorum, and we don't push anything if no operator belonged to that quorum. This also means that if 2 operators had a net-zero impact on the total, we'd push an update that ultimately didn't change the stake.

The current PR does things a little differently, and just checks each quorum to see if the delta is nonzero before pushing an update. This is similar to the current behavior in that if there were no operators belonging to the quorum, nothing gets pushed.

It differs because it also won't push if the delta was 0.

My question is - what do we want here?
- we can keep this behavior and just not push updates if there was no change
- or, we can change this slightly to also record net-zero updates if the last update is not the current block

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -586,15 +586,16 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     ) external view returns (uint32[] memory) {
         uint32[] memory indices = new uint32[](operatorIds.length);
         for (uint256 i = 0; i < operatorIds.length; i++) {
-            uint32 length = uint32(_operatorIdToQuorumBitmapHistory[operatorIds[i]].length);
-            for (uint32 j = 0; j < length; j++) {
+            uint length = _operatorIdToQuorumBitmapHistory[operatorIds[i]].length;
```
gpsanant, line 589: we explicitly declare uint256 usually

src/StakeRegistry.sol
```diff
@@ -283,71 +251,111 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+    function _calculateDelta(uint96 prev, uint96 cur) internal pure returns (int256) {
+        if (cur >= prev) {
+            return int256(uint256(cur - prev));
+        } else {
+            return -int256(uint256(prev - cur));
+        }
+    }
+
+    /// @notice Adds or subtracts delta from value, according to its sign
+    function _applyDelta(uint96 value, int256 delta) internal pure returns (uint96) {
```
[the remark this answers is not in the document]
gpsanant: or int104 or whatev

src/StakeRegistry.sol
```diff
@@ -434,10 +442,10 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
                 _totalStakeHistory[quorumNumber][0].updateBlockNumber <= blockNumber,
                 "StakeRegistry.getTotalStakeIndicesByQuorumNumbersAtBlockNumber: quorum has no stake history at blockNumber"
             );
-            uint32 length = uint32(_totalStakeHistory[quorumNumber].length);
-            for (uint32 j = 0; j < length; j++) {
+            uint length = _totalStakeHistory[quorumNumber].length;
```
gpsanant, line 445: uint256

src/StakeRegistry.sol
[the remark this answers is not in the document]
wadealexc: I switched to i256 delta because I was running into overflow issues when calculating deltas using 0 and `type(uint96).max`

src/StakeRegistry.sol
```diff
@@ -283,71 +251,111 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+        _totalStakeHistory[quorumNumber].push(OperatorStakeUpdate({
+            updateBlockNumber: uint32(block.number),
+            nextUpdateBlockNumber: 0,
+            stake: _applyDelta(prevStake, stakeDelta)
+        }));
+    }
+
+    /// @notice Returns the change between a previous and current value as a signed int
+    function _calculateDelta(uint96 prev, uint96 cur) internal pure returns (int256) {
+        if (cur >= prev) {
```
gpsanant, line 345: `return int(cur) - int(prev)` should work?

wadealexc: i knew there was a better way to do this! good catch.

src/StakeRegistry.sol
gpsanant: not pushing during net zero, should be fine :+1:

src/StakeRegistry.sol
gpsanant: sure, that seems reasonable to me

src/StakeRegistry.sol
gpsanant: that would be a great change! i often need to take a second to remember how to handle the case of many updates in the same block

src/StakeRegistry.sol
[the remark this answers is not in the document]
gpsanant: ok, if in memory, not a big deal

wadealexc:
Well, unsurprisingly the last update broke several tests. We're doing a lot of tests that do a ton of actions and then rely on the previously-linear stake update history to check things. I wasn't able to get them all rewritten today.

src/StakeRegistry.sol
```diff
@@ -253,91 +246,117 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
         bytes32 operatorId,
         uint8 quorumNumber
-    ) internal returns (int256, uint96) {
+    ) internal returns (int256 delta, bool hasMinimumStake) {
         /**
          * Get the operator's current stake for the quorum. If their stake
          * is below the quorum's threshold, set their stake to 0
          */
         uint96 currentStake = weightOfOperatorForQuorum(quorumNumber, operator);
         if (currentStake < minimumStakeForQuorum[quorumNumber]) {
```
gpsanant, line 262: this is not super clean... i think there's an argument that operators should be deregistered from this quorum in this case, but later...

wadealexc: They are. There are 2 cases:
- when an operator tries to register and doesn't have the minimum stake, `registerOperator` reverts
- when an operator deregisters, this returns true, but also that's what we're doing

Or are you suggesting something else?

PR #28, opened 2023-10-26
refactor: quorum creation is the divine right of the registry coordinator

wadealexc:
WIP PR. So far:
* Moved `createQuorum` to the registry coordinator, and added quorum creation logic for each registry contract


src/IndexRegistry.sol
```diff
@@ -48,9 +48,12 @@ contract IndexRegistry is IndexRegistryStorage {
...
 
-            //this is the would-be index of the operator being registered, the total number of operators for that quorum (which is last index + 1)
-            uint256 quorumHistoryLength = _totalOperatorsHistory[quorumNumber].length;
-            uint32 numOperators = quorumHistoryLength > 0 ? _totalOperatorsHistory[quorumNumber][quorumHistoryLength - 1].numOperators : 0;
+            // Check that the quorum exists
+            uint256 historyLength = _totalOperatorsHistory[quorumNumber].length;
+            require(historyLength != 0, "IndexRegistry.registerOperator: quorum does not exist");
+
+            // Get the number of operators currently in the quorum, which will be operator's new index
+            uint32 numOperators = historyLength > 0 ? _totalOperatorsHistory[quorumNumber][historyLength - 1].numOperators : 0;
```
gpsanant, line 56: no need for the > 0 check here if adding require

src/interfaces/IStakeRegistry.sol
```diff
@@ -1,4 +1,160 @@
...
+        // the block number at which the stake amounts were updated and stored
+        uint32 updateBlockNumber;
+        // the block number at which the *next update* occurred.
+        /// @notice This entry has the value **0** until another update takes place.
+        uint32 nextUpdateBlockNumber;
+        // stake weight for the quorum
+        uint96 stake;
+    }
+
+    // EVENTS
```
gpsanant, line 32: 2 event sections should be made 1

src/StakeRegistry.sol
```diff
@@ -203,15 +187,85 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
         }
     }
 
-    /*******************************************************************************
-                    EXTERNAL FUNCTIONS - SERVICE MANAGER OWNER
-    *******************************************************************************/
+    /// @notice Create a new quorum and add the strategies and their associated weights to the quorum.
+    function createQuorum(
```
gpsanant, line 191: this needs to push something to total stake history in order for the quorumExists condition to be true

src/BLSPubkeyRegistry.sol
```diff
@@ -103,28 +117,32 @@ contract BLSPubkeyRegistry is BLSPubkeyRegistryStorage {
...
+                nextUpdateBlockNumber: 0
+            }));
 
             unchecked {
                 ++i;
             }
         }
     }
 
+    // TODO - should this fail if apkUpdate.apkHash == 0? This will be the case for the first entry in each quorum
```
ChaoticWalrus, line 145: good question. relatedly, should we actually push the hash of the zero point in the first update, rather than zero per se?
I think this probably _shouldn't_ revert in this case, but am not strongly opinionated

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -99,25 +108,30 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     function initialize(
         address _churnApprover,
         address _ejector,
-        OperatorSetParam[] memory _operatorSetParams,
         IPauserRegistry _pauserRegistry,
-        uint256 _initialPausedStatus
+        uint256 _initialPausedStatus,
+        OperatorSetParam[] memory _operatorSetParams,
+        uint96[] memory _minimumStakes,
+        IStakeRegistry.StrategyAndWeightingMultiplier[][] memory _strategyParams
```
ChaoticWalrus, line 115: should this be called `_initialQuorumParams` or something?

ChaoticWalrus: ehh actually the other things are related enough to quorums, I think it makes at least as much sense as-is.

src/BLSPubkeyRegistry.sol
wadealexc: question for @gpsanant

src/IndexRegistry.sol
```diff
@@ -119,10 +134,10 @@ contract IndexRegistry is IndexRegistryStorage {
...
      */ 
     function _updateOperatorIdToIndexHistory(bytes32 operatorId, uint8 quorumNumber, uint32 index) internal {
-        OperatorUpdate memory latestOperatorUpdate;
-        latestOperatorUpdate.operatorId = operatorId;
-        latestOperatorUpdate.fromBlockNumber = uint32(block.number);
-        _indexToOperatorIdHistory[quorumNumber][index].push(latestOperatorUpdate);
+        _indexToOperatorIdHistory[quorumNumber][index].push(OperatorUpdate({
+            operatorId: operatorId,
+            fromBlockNumber: uint32(block.number)
+        }));
```
ChaoticWalrus, line 140: it's weird that elsewhere we use 24 bytes (as in the BLSPubkeyRegistry) for the equivalent of operatorId, and here we use 32 bytes. Not sure why this line in particular made me notice this, but it's just...a bit odd?
should operatorIds perhaps be shorter, so we can do the same storage packing everywhere?

src/StakeRegistry.sol
```diff
@@ -78,9 +56,14 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
         // for each quorum, loop through operators and see if they are a part of the quorum
         // if they are, get their new weight and update their individual stake history and the
         // quorum's total stake history accordingly
+        uint8 quorumCount = registryCoordinator.quorumCount();
         for (uint8 quorumNumber = 0; quorumNumber < quorumCount; ) {
             int256 totalStakeDelta;
 
+            // TODO - not a huge fan of this dependency on the reg coord, but i do prefer this
+            //        over the stakereg also keeping its own count.
```
ChaoticWalrus, line 64: I think this makes sense. it's a bit ugly, but in my mind the registries are basically all "puppets" of the RegistryCoordinator, so calling into the coordinator to fetch a bit of data is fine.
I wouldn't be super opposed to just keeping count in the StakeRegistry either, as it'd be more efficient -- we could just increment a `quorumCount` variable in the `initializeQuorum()` function? Then the two can only be out-of-sync if the RegistryCoordinator either fails to call `initializeQuorum` appropriately or calls it inappropriately.

src/StakeRegistry.sol
```diff
@@ -203,15 +186,91 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+        uint96 minimumStake
+    ) public virtual onlyServiceManagerOwner quorumExists(quorumNumber) {
         _setMinimumStakeForQuorum(quorumNumber, minimumStake);
     }
 
+    /** 
+     * @notice Adds strategies and weights to the quorum
+     * @dev Checks to make sure that the *same* strategy cannot be added multiple times (checks against both against existing and new strategies).
+     * @dev This function has no check to make sure that the strategies for a single quorum have the same underlying asset. This is a concious choice,
+     * since a middleware may want, e.g., a stablecoin quorum that accepts USDC, USDT, DAI, etc. as underlying assets and trades them as "equivalent".
```
ChaoticWalrus, line 217: I realize these probably weren't part of this PR (it probably just moves the comments?) but:
typo: 'concious' => 'conscious'
and I think 'trades' should be 'treats'

src/StakeRegistry.sol
```diff
@@ -203,15 +186,91 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+
+        for (uint256 i = 0; i < toRemoveLength; i++) {
+            emit StrategyRemovedFromQuorum(quorumNumber, strategyParams[indicesToRemove[i]].strategy);
+            emit StrategyMultiplierUpdated(quorumNumber, strategyParams[indicesToRemove[i]].strategy, 0);
+
+            // Replace index to remove with the last item in the list, then pop the last item
+            strategyParams[indicesToRemove[i]] = strategyParams[strategyParams.length - 1];
+            strategyParams.pop();
+        }
+    }
```
ChaoticWalrus, line 248: @wadealexc thoughts on the safety of this function?
it previously had a redundant parameter which it checked against, to ensure that the desired strategies were removed

ChaoticWalrus (approved):
Feels like a good change overall.
I made some smaller comments but don't think I have any major points.
Am in agreement with the small open items noted by @gpsanant

src/StakeRegistry.sol
wadealexc: This is marked TODO because I was intending a future PR to move `updateStakes` into the registry coordinator. I'm working on that locally and will have something to share next week

src/StakeRegistry.sol
wadealexc: What redundant parameter? Here's the master branch: https://github.com/Layr-Labs/eigenlayer-middleware/blob/master/src/VoteWeigherBase.sol#L62

wadealexc: I havent given any thought to the safety of the method... I think that'll come when we start working on integration tests. Essentially we should be testing that stake updates occur as expected even if we interweave strategy modifications in between updates.

src/StakeRegistry.sol
ChaoticWalrus: ahhh OK that works!

src/StakeRegistry.sol
ChaoticWalrus: sorry, wasn't related to this PR, probably should have asked elsewhere, but here's a link to the previous behavior I was thinking of https://github.com/Layr-Labs/eigenlayer-contracts/blob/bd211f12fe8ed84f691ca44db3f812412defcb1e/src/contracts/middleware/VoteWeigherBase.sol#L134-L138

this could be described as an "idiot-check", but I do have a feeling it'd be easy to call this function "incorrectly" (or at least, in a way that doesn't have the desired outcome) with the way it works now

src/BLSPubkeyRegistry.sol
wadealexc: so to confirm, the change is to push the zero hash for the first update, and have this method fail if `apkHash == 0`?

Should it also fail if `apkHash == ZERO_HASH`?

wadealexc: so to confirm, the change is to push the zero hash for the first update, and have this method fail if `apkHash == 0`?

Should it also fail if `apkHash == ZERO_HASH`?

PR #29, opened 2023-10-26
Migrate interfaces from core repo

ChaoticWalrus:
Sister PR to https://github.com/Layr-Labs/eigenlayer-contracts/pull/294
~~TODO: update the submodule commit that this repo is using. I got this compiling with tests running fine locally though!~~
Edit: looks like this is compiling + passing CI, am marking as ready for review

### DIFF
--- a/src/interfaces/IServiceManager.sol
+++ b/src/interfaces/IServiceManager.sol
@@ -2,7 +2,7 @@
 pragma solidity >=0.5.0;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import {Slasher} from "eigenlayer-contracts/src/contracts/core/Slasher.sol";
+import {ISlasher} from "eigenlayer-contracts/src/contracts/interfaces/ISlasher.sol";
 
 /**
  * @title Interface for a `ServiceManager`-type contract.
--- a/src/interfaces/IServiceManager.sol
+++ b/src/interfaces/IServiceManager.sol
@@ -1,4 +1,39 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.5.0;
 
-import "eigenlayer-contracts/src/contracts/interfaces/IServiceManager.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {Slasher} from "eigenlayer-contracts/src/contracts/core/Slasher.sol";
+
+/**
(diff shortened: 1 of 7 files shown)


src/interfaces/IServiceManager.sol
```diff
@@ -1,4 +1,39 @@
...
+
+    /// @notice proxy call to the slasher, recording a stake update
+    function recordStakeUpdate(address operator, uint32 updateBlock, uint32 serveUntilBlock, uint256 prevElement) external;
+
+    /// @notice proxy call to the slasher, recording a final stake update (on operator deregistration)
+    function recordLastStakeUpdateAndRevokeSlashingAbility(address operator, uint32 serveUntilBlock) external;
+
+    /// @notice Returns the latest block until which operators must serve (could be in the past or future).
+    /// @dev this should be called and the response passed to the recordStakeUpdate functionss' serveUntilBlock parameter
+    function latestServeUntilBlock() external view returns (uint32);
```
samlaf, line 35: Can we remove all of these from the interface? Slasher interface is likely going to change, so there's no point in having these. We still want the AVS teams to use this interface for the upcoming M2 testnet release, so need to make an interface that makes sense.

samlaf (commented):
Can we remove all stakeUpdate related functionality from the IServiceManager interface? We can keep the `freezeOperator` call since that one is unlikely to change, but all the rest feels like will most likely change.

Will then need to update the ServiceManagerBase to remove all of this recordStakeUpdate proxy calls. We want this to release incredible-squaring and have a consistent story to tell when trying to onboard AVSs onto our testnet.

src/interfaces/IServiceManager.sol
ChaoticWalrus: I'm OK with that but would prefer grouping it into a separate PR. This PR is really just aimed at migrating the existing interfaces.

samlaf: Fine with me. Can we get this in tomorrow?

ChaoticWalrus: > Can we get this in tomorrow?

I'm game 👍

stevennevins (approved):

src/interfaces/IServiceManager.sol
stevennevins: https://github.com/Layr-Labs/eigenlayer-middleware/issues/33

PR #31, opened 2023-10-26
FEAT: update index reg and propagate changes. need to fix indexreg tests

gpsanant:
Simplify index registry.

The aim to to get rid of the operators to swap param.

Some downsides: 
- operatorUpdate is now 2 storage slots
- new map for current operator index for quorums

However, the global operator list is removed, so that reduces a fair amount of storage

### DIFF
--- a/src/IndexRegistry.sol
+++ b/src/IndexRegistry.sol
@@ -42,8 +42,6 @@ contract IndexRegistry is IndexRegistryStorage {
         bytes calldata quorumNumbers
     ) public virtual onlyRegistryCoordinator returns(uint32[] memory) {
         uint32[] memory numOperatorsPerQuorum = new uint32[](quorumNumbers.length);
-        //add operator to operatorList
-        globalOperatorList.push(operatorId);
 
         for (uint256 i = 0; i < quorumNumbers.length; i++) {
             uint8 quorumNumber = uint8(quorumNumbers[i]);
@@ -70,8 +68,6 @@ contract IndexRegistry is IndexRegistryStorage {
      * @notice Deregisters the operator with the specified `operatorId` for the quorums specified by `quorumNumbers`.
      * @param operatorId is the id of the operator that is being deregistered
      * @param quorumNumbers is the quorum numbers the operator is deregistered for
-     * @param operatorIdsToSwap is the list of operatorIds that have the largest indexes in each of the `quorumNumbers`
-     * they will be swapped with the operator's current index when the operator is removed from the list
      * @dev access restricted to the RegistryCoordinator
      * @dev Preconditions (these are assumed, not validated in this contract):
      *         1) `quorumNumbers` has no duplicates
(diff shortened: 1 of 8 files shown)


src/IndexRegistryStorage.sol
```diff
@@ -13,17 +13,19 @@ import "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
...
 
     /// @notice list of all operators ever registered, may include duplicates. used to avoid running an indexer on nodes
     bytes32[] public globalOperatorList;
 
-    /// @notice mapping of operatorId => quorumNumber => index history of that operator
-    mapping(bytes32 => mapping(uint8 => OperatorIndexUpdate[])) internal _operatorIdToIndexHistory;
+    /// @notice mapping of quorumNumber => operator id => current index
+    mapping(uint8 => mapping(bytes32 => uint32)) public operatorIdToIndex;
+    /// @notice mapping of quorumNumber => index => operator id history for that index
+    mapping(uint8 => mapping(uint32 => OperatorUpdate[])) internal _indexToOperatorIdHistory;
```
ChaoticWalrus, line 28: oooohhhh OK nice, this is the crux of the change, isn't it? structuring the storage different so you can look things up from the index.
but it means not being able to get operator IDs' historic indices easily -- I take it that's OK? Is the main existing use for these type of lookups the swap operations (where only current index matters)?

src/IndexRegistry.sol
```diff
@@ -147,30 +138,24 @@ contract IndexRegistry is IndexRegistryStorage {
...
-        bytes32 operatorIdToSwap
+        uint32 indexOfOperatorToRemove 
     ) internal {   
-        uint32 operatorIdToSwapIndex = _operatorIdToIndexHistory[operatorIdToSwap][quorumNumber][_operatorIdToIndexHistory[operatorIdToSwap][quorumNumber].length - 1].index;
-        require(
-            _totalOperatorsHistory[quorumNumber][_totalOperatorsHistory[quorumNumber].length - 1].numOperators - 1 == operatorIdToSwapIndex, 
-            "IndexRegistry._processOperatorRemoval: operatorIdToSwap is not the last operator in the quorum"
-        );
-
+        uint32 currentNumOperators = _totalOperatorsHistory[quorumNumber][_totalOperatorsHistory[quorumNumber].length - 1].numOperators;
```
ChaoticWalrus, line 143: neat! might want to make a getter/util like `numberOfOperators(uint8 quorumNumber)` ?

ChaoticWalrus (commented):
Big-brain PR, looking really good.

src/IndexRegistryStorage.sol
gpsanant: the idea is that we only need that for proofs of custody, and we'll have more sophisticated indexing support by then

src/IndexRegistry.sol
gpsanant: good for clarity: https://github.com/Layr-Labs/eigenlayer-middleware/pull/31/commits/657c50c335526c4ca0cd0671c2bed90a1061ce2c

ChaoticWalrus: > good for clarity: [657c50c](https://github.com/Layr-Labs/eigenlayer-middleware/commit/657c50c335526c4ca0cd0671c2bed90a1061ce2c)

wait, why not just make the existing 'view' method public?
I guess it's fine this way though 🤷

gpsanant: i'm to sauce idk if we are cool with public functions?

ChaoticWalrus: > i'm to sauce idk if we are cool with public functions?

IMO:
public 'view' functions are fine
public state-modifying functions are often potential landmines (it can be hard to tell when access control is correct on them, in particular)

ChaoticWalrus (approved):
Seems like a positive move.
There's a couple minor threads open which we can still work out before merging, but structurally this looks good and I think is a solid decision (IMO the 'ids to swap' param is kinda cursed in general, so I'm glad to see it die).

This mildly changes interfaces so just want to flag that in case merging it to master will result in a lot of work for other teams?

src/IndexRegistryStorage.sol
ChaoticWalrus: > the idea is that we only need that for proofs of custody, and we'll have more sophisticated indexing support by then

ah OK but we can still use off-chain indexing and prove inputs are correct for that. Seems fine then.

src/IndexRegistry.sol
```diff
@@ -130,15 +119,17 @@ contract IndexRegistry is IndexRegistryStorage {
...
+        latestOperatorUpdate.fromBlockNumber = uint32(block.number);
+        _indexToOperatorIdHistory[quorumNumber][index].push(latestOperatorUpdate);
+
+        operatorIdToIndex[quorumNumber][operatorId] = index;
 
         emit QuorumIndexUpdate(operatorId, quorumNumber, index);
     }
 
-    /**
+    /**v
```
wadealexc, line 132: v

gpsanant: wut

wadealexc: v

ChaoticWalrus: ahhh, I see!
v

gpsanant: v v
____

Issue #33, opened 2023-10-27
Chore: Remove Stake Update Functionality from ServiceManager


stevennevins:
Can we remove all stakeUpdate related functionality from the IServiceManager interface? We can keep the `freezeOperator` call since that one is unlikely to change, but all the rest feels like will most likely change.

Will then need to update the ServiceManagerBase to remove all of this recordStakeUpdate proxy calls. We want this to release incredible-squaring and have a consistent story to tell when trying to onboard AVSs onto our testnet.

_Originally posted by @samlaf in https://github.com/Layr-Labs/eigenlayer-middleware/pull/29#pullrequestreview-1700879020_

PR #34, opened 2023-10-27
Chore: remove unneeded interface methods

stevennevins:
https://github.com/Layr-Labs/eigenlayer-middleware/issues/33

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -9,9 +9,7 @@ import "src/interfaces/IServiceManager.sol";
 
 import "eigenlayer-contracts/src/contracts/permissions/Pausable.sol";
 import "eigenlayer-contracts/src/contracts/interfaces/IPauserRegistry.sol";
-import "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
 import "eigenlayer-contracts/src/contracts/interfaces/ISlasher.sol";
-import "eigenlayer-contracts/src/contracts/interfaces/IStrategyManager.sol";
 
 /**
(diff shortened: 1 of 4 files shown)


ChaoticWalrus (approved):
Works for me.
We aren't using these functions in M2 so should be fine to delete; they can be added back later.

samlaf:
Thanks @stevennevins ! Are you planning to update the ServiceManagerBase in another PR? If not, can we include it here (it was part of the issue #33 )

samlaf:
Thanks for the quick changes @ChaoticWalrus !!
As we've talked offline, question now is whether this ServiceManagerBase is even worth keeping at all. Don't have enough contracts experience to make a good decision here, so will let you guys decide, but ideal solution would be *actually* useful for AVS devs, without getting more random requirements in their way.

Perhaps we can merge this PR and discuss this issue separately?

stevennevins:
> Thanks for the quick changes @ChaoticWalrus !! As we've talked offline, question now is whether this ServiceManagerBase is even worth keeping at all. Don't have enough contracts experience to make a good decision here, so will let you guys decide, but ideal solution would be _actually_ useful for AVS devs, without getting more random requirements in their way.
> 
> Perhaps we can merge this PR and discuss this issue separately?

I'm pro removing anything that seems like bloat and isn't immediately useful

ChaoticWalrus:
> I'm pro removing anything that seems like bloat and isn't immediately useful

After reviewing ServiceManagerBase, I think it's pretty close to an empty shell at the moment. Would be fine deleting it, especially if AVS team isn't finding it useful.

Issue #27, opened 2023-10-26
Add explanations above commented out slasher integration code


samlaf:
We decided to leave slasher integration out of the upcoming M2 testnet deployment.
However, we have only reflected that choice by commenting out all the places where slasher calls are made.
Some of them have a TODO comment, for eg
https://github.com/Layr-Labs/eigenlayer-middleware/blob/f9c6508c85d971de133498aef29b454faba70d9c/src/StakeRegistry.sol#L119

Others are simply commented without comments, such as https://github.com/Layr-Labs/eigenlayer-middleware/blob/f9c6508c85d971de133498aef29b454faba70d9c/src/BLSRegistryCoordinatorWithIndices.sol#L381

Could we be consistent about these messages. Would be nice to have a single standardized message. This would help code readability, would make it easier to find all these instances when we do want to reenable the slasher, and would make it easier for the AVS team to communicate with external AVS devs the choices that are currently being made.

stevennevins:
https://github.com/Layr-Labs/eigenlayer-middleware/pull/36

samlaf:
@stevennevins #36 was merged, I think we can close this? 
If you add "Closes #PR_NUMBER" somewhere in the description of the PR, it'll automatically close linked issues.

PR #36, opened 2023-10-27
docs: Consistently Indicate Slasher Under Development and Inactive

stevennevins:
per Sam's feedback https://github.com/Layr-Labs/eigenlayer-middleware/issues/27


src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -378,6 +378,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
         BN254.G1Point memory pubkey, 
         string memory socket
     ) internal virtual returns(uint32[] memory) {
+        // TODO after slashing: revert registration when operator is slashable in the EigenLayer Slasher
```
ChaoticWalrus, line 381: nit: this should be when they are *not* slashable (but the interface might change so... 🤷 )

ChaoticWalrus (approved):
One small nit on wording, otherwise looks good-to-go

src/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: I'm fine with that 👍

samlaf: Wait interface might change but we will always require being opted into slashing to be able to register right?
Feels like we should remove the commented code below (since the interface will change), but not this comment?

ChaoticWalrus: > Wait interface might change but we will always require being opted into slashing to be able to register right? Feels like we should remove the commented code below (since the interface will change), but not this comment?

tbh I'm fine either way. The communication about what being "opted into slashing" means might change, but having a TODO left in the code seems fine

samlaf:
@stevennevins @ChaoticWalrus just realized, can we also add a note above freezeOperator() in IServiceManager.sol saying that this is only a temporary placeholder, and shouldnt actually be implemented to call the slasher for now.

PR #37, opened 2023-10-27
Gas Analysis: StakeRegistry.updateStakes() setup as a cronjob

8sunyuan:
Description: 
We want to get some gas cost estimates for performing a weekly running cronjob that performs stake updates for 
operators. Assuming ~200 operators, and varying quorums, this action would probably have to be batched into multiple update calls.

### DIFF
--- a/test/unit/StakeRegistryUnit.t.sol
+++ b/test/unit/StakeRegistryUnit.t.sol
@@ -540,6 +540,7 @@ contract StakeRegistryUnitTests is Test {
         }
     }
 
+    /// forge-config: default.fuzz.runs = 5
     function testUpdateStakes_200Operators(
         address[200] calldata operators,
         uint256[200] calldata pseudoRandomNumbers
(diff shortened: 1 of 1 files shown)


test/unit/StakeRegistryUnit.t.sol
```diff
@@ -614,3 +617,378 @@ contract StakeRegistryUnitTests is Test {
...
+
+            uint32 intialBlockNumber = 100;
+            cheats.roll(intialBlockNumber);
+
+            bytes memory quorumNumbers = BitmapUtils.bitmapToBytesArray(quorumBitmap);
+            for (uint j = 0; j < stakesForQuorum.length; j++) {
+                stakeRegistry.setOperatorWeight(uint8(quorumNumbers[j]), defaultOperator, stakesForQuorum[j] + 1);
+            }
+
+            updateOperators[i] = operators[i];
```
ChaoticWalrus, line 972: this looks like it's repeatedly setting the weight just for the defaultOperator?

ChaoticWalrus: oh, but `defaultOperator = operators[i];` above. never mind, just confusing naming

8sunyuan: Yeah, carried over some of the logic from existing tests so not the cleanest at the moment.

PR #39, opened 2023-10-30
chore: add missing config for avs deploy

stevennevins:
Add back the missing config file for AVSContractsDeploy.s.sol

Simulate with:
```
forge script script/AVSContractsDeploy.s.sol
```


ChaoticWalrus (approved):
Seems fine.
This appears to be for Goerli + can probably have a bit of cleanup done, but if we're just adding it back after deletion I'm OK with this as-is.

samlaf:
@ChaoticWalrus @stevennevins was just looking for this script, thanks!
Btw, can we make it a habit to rebase on top of master instead of merging master into our branches. Doesn't make that much of a difference for small PRs like this, but it keeps the history much cleaner, which is really useful when we have long standing branches like multiquorums and avs-unstable. We'll probably end up with a similar tangled mess here if we keep merging master into the `m2-mainnet` and `alex/refactor-quorums` branches.

samlaf:
Oh also just realized there's a big misconfiguration... the `AVSContractsDeploy.s.sol` file actually contains the `EigenLayerDeploy` contract... how are you guys deploying the AVS contracts??

stevennevins:
> @ChaoticWalrus @stevennevins was just looking for this script, thanks! Btw, can we make it a habit to rebase on top of master instead of merging master into our branches. Doesn't make that much of a difference for small PRs like this, but it keeps the history much cleaner, which is really useful when we have long standing branches like multiquorums and avs-unstable. We'll probably end up with a similar tangled mess here if we keep merging master into the `m2-mainnet` and `alex/refactor-quorums` branches.

Yeah, I generally agree here; it seems like that's going to be the plan for the `m2-mainnet` branch

stevennevins:
> Oh also just realized there's a big misconfiguration... the `AVSContractsDeploy.s.sol` file actually contains the `EigenLayerDeploy` contract... how are you guys deploying the AVS contracts??

Yeah, this could be improved. I'm just trying to fix the broken script that currently exists with this PR.

PR #43, opened 2023-10-31
refactor: simplify naming and logic in registries and registry coordinator

wadealexc:
Motivation:
* fix word soup variable and function naming
* take advantage of new quorum existence invariant due to `RegistryCoordinator.createQuorum` and its associated `initializeQuorum` methods in each registry
* registries each keep "histories" of updates. The status quo is that every time anything changes, we push a new history update. this means that when updates happen multiple times in the same block, the history contains multiple updates from the same block number, making it harder to determine what the canonical state at a given block number is.

This PR:
* Replaces logic that handles "0-length histories" with simple checks, since we now have an invariant that initialized quorums have nonzero histories (see any registry's `initializeQuorum` method)
* changes registry histories to update their "most recent entry" rather than pushing a new entry, if the entry was made in the current block
* Renames state variables in the registry coordinator and registry contracts to be less wordy

**For reviewers, this order may work best:**
1. Review `IndexRegistry` changes in the first commit, which contains the functional changes mentioned above, as well as readability/logical simplification. This is the worst commit to review, sorry! The diff looks intimidating, so I'd recommend opening the new/old IndexRegistry files side by side and comparing like that. :pray: 
2. The rest of the commits are style/refactor commits. I'd recommend skimming the renaming changes below and only dipping into the commits themselves if you feel like it.

Renaming changes:
* `IndexRegistry` state vars:
  * `operatorIdToIndex` -> `currentOperatorIndex`
  * `_totalOperatorsHistory` -> `_operatorCountHistory`
  * `_indexToOperatorIdHistory` -> `_indexHistory`
* `IndexRegistry` functions:
  * `_getTotalOperatorsForQuorumAtBlockNumber` -> `_operatorCountAtBlockNumber`
  * `_getOperatorIdAtIndexForQuorumAtBlockNumber` -> `_operatorIdForIndexAtBlockNumber`
  * `getOperatorIndexUpdateOfIndexForQuorumAtIndex` -> `getOperatorUpdateAtIndex`
  * `getTotalOperatorsForQuorumAtBlockNumberByIndex` -> `getTotalOperatorsForIndexAtBlockNumber`
  * `getOperatorListForQuorumAtBlockNumber` -> `getOperatorListAtBlockNumber`
* `StakeRegistry` state vars:
  * `operatorIdToStakeHistory` -> `operatorStakeHistory`
  * `strategiesConsideredAndMultipliers` -> `strategyParams`
* `StakeRegistry` structs:
  * `StrategyAndWeightingMultiplier` -> `StrategyParams`
* `StakeRegistry` functions:
  * `strategiesConsideredAndMultipliersLength` -> `strategyParamsLength`
  * `strategyAndWeightingMultiplierForQuorumByIndex` -> `strategyParamsByIndex`
  * `getOperatorIdToStakeHistory` -> `getOperatorStakeHistory`
  * `getLengthOfTotalStakeHistoryForQuorum` -> `getTotalStakeHistoryLength`
  * `getTotalStakeUpdateForQuorumFromIndex` -> `getTotalStakeUpdateAtIndex`
  * `getStakeUpdateIndexForOperatorIdForQuorumAtBlockNumber` -> `getStakeUpdateIndexForOperatorAtBlockNumber`
  * `getTotalStakeIndicesByQuorumNumbersAtBlockNumber` -> `getTotalStakeIndicesAtBlockNumber`
  * `getStakeUpdateForQuorumFromOperatorIdAndIndex` -> `getStakeUpdateForOperatorAtIndex`
  * `getStakeForQuorumAtBlockNumberFromOperatorIdAndIndex` -> `getOperatorStakeAtBlockNumberAndIndex`
  * `getStakeForOperatorIdForQuorumAtBlockNumber` -> `getOperatorStakeAtBlockNumber`
  * `getLengthOfOperatorIdStakeHistoryForQuorum` -> `getOperatorStakeHistoryLength`
* `BLSRegistryCoordinatorWithIndices` state vars:
  * `_quorumOperatorSetParams` -> `_quorumParams`
  * `_operatorIdToQuorumBitmapHistory` -> `_operatorBitmapHistory`
  * `_operators` -> `_operatorInfo`
* `BLSRegistryCoordinatorWithIndices` state modifying functions:
  * `registerOperatorWithCoordinator` -> `registerOperator`
  * `registerOperatorWithCoordinator` -> `registerOperatorWithChurn`
  * `deregisterOperatorWithCoordinator` -> `deregisterOperator`
  * `ejectOperatorFromCoordinator` -> `ejectOperator`
* `BLSRegistryCoordinatorWithIndices` view functions:
  * `getQuorumBitmapIndicesByOperatorIdsAtBlockNumber` -> `getQuorumBitmapIndicesAtBlockNumber`
  * `getQuorumBitmapByOperatorIdAtBlockNumberByIndex` -> `getQuorumBitmapAtBlockNumberByIndex`
  * `getQuorumBitmapUpdateByOperatorIdByIndex` -> `getQuorumBitmapUpdateByIndex`
  * `getCurrentQuorumBitmapByOperatorId` -> `getCurrentQuorumBitmap`
  * `getQuorumBitmapUpdateByOperatorIdLength` -> `getQuorumBitmapHistoryLength`

`IndexRegistry` changes:
* Refactored index registry history updates to have a more readable register/deregister flow:
  * `registerOperator` is now `_increaseOperatorCount` + `_assignOperatorToIndex`. Functional changes:
    * If the most recent `QuorumUpdate/OperatorUpdate` is from this block, we update the last entry rather than pushing a new one.
  * `deregisterOperator` is now `_decreaseOperatorCount` + `_popLastOperator` + `_assignOperatorToIndex`. Functional changes:
    * As above, histories are updated rather than pushed if the last entry was from the current block.
    * Previously, we assigned `currentOperatorIndex[quorum][OPERATOR_DOES_NOT_EXIST_ID]` to the removed index, and emitted an event for this update. This logic has been removed, since the `currentOperatorIndex` of the null id doesn't matter. See [previous logic](https://github.com/Layr-Labs/eigenlayer-middleware/blob/alex/refactor-quorums/src/IndexRegistry.sol#L169-L174) vs [new logic](https://github.com/Layr-Labs/eigenlayer-middleware/blob/aa1d513253c5a032d68e398a05b86479812bc68b/src/IndexRegistry.sol#L175)

### DIFF
--- a/src/BLSRegistryCoordinatorWithIndices.sol
+++ b/src/BLSRegistryCoordinatorWithIndices.sol
@@ -145,17 +145,24 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
      * @param pubkey is the BLS public key of the operator
      * @param socket is the socket of the operator
      */
-    function registerOperatorWithCoordinator(
+    function registerOperator(
         bytes calldata quorumNumbers,
         BN254.G1Point memory pubkey,
         string calldata socket
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
-        _registerOperatorWithCoordinatorAndNoOverfilledQuorums({
+        uint32[] memory numOperatorsPerQuorum = _registerOperator({
             operator: msg.sender, 
             quorumNumbers: quorumNumbers, 
             pubkey: pubkey, 
             socket: socket
         });
+
+        for (uint256 i = 0; i < numOperatorsPerQuorum.length; i++) {
+            require(
+                numOperatorsPerQuorum[i] <= _quorumParams[uint8(quorumNumbers[i])].maxOperatorCount,
+                "BLSRegistryCoordinatorWithIndices.registerOperator: quorum is overfilled"
+            );
+        }
     }
 
     /**
@@ -168,7 +175,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
      * and ids of the operators to swap with the kicked operator. 
      * @param churnApproverSignature is the signature of the churnApprover on the operator kick params
      */
-    function registerOperatorWithCoordinator(
+    function registerOperatorWithChurn(
         bytes calldata quorumNumbers, 
         BN254.G1Point memory pubkey,
         string calldata socket,
@@ -176,7 +183,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
         SignatureWithSaltAndExpiry memory churnApproverSignature
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
         // register the operator
-        uint32[] memory numOperatorsPerQuorum = _registerOperatorWithCoordinator({
+        uint32[] memory numOperatorsPerQuorum = _registerOperator({
             operator: msg.sender,
             quorumNumbers: quorumNumbers,
             pubkey: pubkey,
@@ -209,7 +216,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
 
                 require(
                     operatorKickParams[operatorToKickParamsIndex].quorumNumber == quorumNumber, 
-                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithCoordinator: quorumNumber not the same as signed"
+                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithChurn: quorumNumber not the same as signed"
                 );
 
                 // get the total stake for the quorum
@@ -221,13 +228,13 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
                 // check the registering operator has more than the kick BIPs of the operator to kick's stake
                 require(
                     registeringOperatorStake > operatorToKickStake * operatorSetParam.kickBIPsOfOperatorStake / BIPS_DENOMINATOR,
-                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithCoordinator: registering operator has less than kickBIPsOfOperatorStake"
+                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithChurn: registering operator has less than kickBIPsOfOperatorStake"
                 );
                 
                 // check the that the operator to kick has less than the kick BIPs of the total stake
                 require(
                     operatorToKickStake < totalStakeForQuorum * operatorSetParam.kickBIPsOfTotalStake / BIPS_DENOMINATOR,
-                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithCoordinator: operator to kick has more than kickBIPSOfTotalStake"
+                    "BLSRegistryCoordinatorWithIndices.registerOperatorWithChurn: operator to kick has more than kickBIPSOfTotalStake"
                 );
 
                 // increment the operatorToKickParamsIndex
@@ -235,7 +242,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
             }
             
             // kick the operator
-            _deregisterOperatorWithCoordinator({
+            _deregisterOperator({
                 operator: operatorKickParams[i].operator,
                 quorumNumbers: quorumNumbers[i:i+1],
                 pubkey: operatorKickParams[i].pubkey 
@@ -248,11 +255,11 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
      * @param quorumNumbers are the bytes representing the quorum numbers that the operator is registered for
      * @param pubkey is the BLS public key of the operator
      */
-    function deregisterOperatorWithCoordinator(
+    function deregisterOperator(
         bytes calldata quorumNumbers,
         BN254.G1Point memory pubkey
     ) external onlyWhenNotPaused(PAUSED_DEREGISTER_OPERATOR) {
-        _deregisterOperatorWithCoordinator({
+        _deregisterOperator({
             operator: msg.sender, 
             quorumNumbers: quorumNumbers, 
             pubkey: pubkey
@@ -264,7 +271,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
      * @param socket is the new socket of the operator
      */
     function updateSocket(string memory socket) external {
-        require(_operatorInfo[msg.sender].status == OperatorStatus.REGISTERED, "BLSRegistryCoordinatorWithIndicies.updateSocket: operator is not registered");
+        require(_operatorInfo[msg.sender].status == OperatorStatus.REGISTERED, "BLSRegistryCoordinatorWithIndices.updateSocket: operator is not registered");
         emit OperatorSocketUpdate(_operatorInfo[msg.sender].operatorId, socket);
     }
 
@@ -278,12 +285,12 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
      * @param quorumNumbers are the quorum numbers to eject the operator from
      * @param pubkey is the BLS public key of the operator
      */
-    function ejectOperatorFromCoordinator(
+    function ejectOperator(
         address operator, 
         bytes calldata quorumNumbers, 
         BN254.G1Point memory pubkey
     ) external onlyEjector {
-        _deregisterOperatorWithCoordinator({
+        _deregisterOperator({
             operator: operator, 
             quorumNumbers: quorumNumbers, 
             pubkey: pubkey
@@ -341,7 +348,7 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     *******************************************************************************/
 
     /// @return numOperatorsPerQuorum is the list of number of operators per quorum in quorumNumberss
-    function _registerOperatorWithCoordinator(
+    function _registerOperator(
         address operator, 
         bytes calldata quorumNumbers, 
         BN254.G1Point memory pubkey, 
@@ -349,8 +356,8 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
     ) internal virtual returns(uint32[] memory) {        
         // Create and validate bitmap from quorumNumbers
         uint256 quorumBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers);
-        require(quorumBitmap <= MAX_QUORUM_BITMAP, "BLSRegistryCoordinatorWithIndices._registerOperatorWithCoordinator: bitmap exceeds max bitmap size");
-        require(quorumBitmap != 0, "BLSRegistryCoordinatorWithIndices._registerOperatorWithCoordinator: bitmap cannot be 0");
+        require(quorumBitmap <= MAX_QUORUM_BITMAP, "BLSRegistryCoordinatorWithIndices._registerOperator: bitmap exceeds max bitmap size");
+        require(quorumBitmap != 0, "BLSRegistryCoordinatorWithIndices._registerOperator: bitmap cannot be 0");
         
         /**
          * Register the operator with the BLSPubkeyRegistry, StakeRegistry, and IndexRegistry. Retrieves:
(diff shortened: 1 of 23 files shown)


src/StakeRegistry.sol
```diff
@@ -386,35 +386,26 @@ contract StakeRegistry is StakeRegistryStorage {
             return;
         }
 
-        uint96 prevStake;
+        // Get our last-recorded stake update
         uint256 historyLength = _totalStakeHistory[quorumNumber].length;
```
gpsanant, line 390: revert if 0

src/IndexRegistry.sol
```diff
@@ -117,121 +126,161 @@ contract IndexRegistry is IndexRegistryStorage {
...
+            _indexHistory[quorumNumber][index].push(OperatorUpdate({
+                operatorId: newOperatorId,
+                fromBlockNumber: uint32(block.number)
+            }));
+        }
+    }
+
+    /// @notice Returns the most recent operator count update for a quorum
+    /// @dev Reverts if the quorum does not exist (history length == 0)
+    function _latestQuorumUpdate(uint8 quorumNumber) internal view returns (QuorumUpdate storage) {
```
gpsanant, line 226: `_getLatestQuorumUpdate`? same below?

src/StakeRegistry.sol
wadealexc: I need to add an explicit comment about this, but this method assumes the quorum exists. all calling locations already perform this check, and will revert higher in the stack if it's caught - so we don't necessarily need an explicit check here.

`updateStakes` doesnt currently check this, but i'm moving that out of this contract anyway.

src/IndexRegistry.sol
```diff
@@ -117,121 +126,170 @@ contract IndexRegistry is IndexRegistryStorage {
...
-     * @param numOperators is the number of operators in the quorum
+     * @notice Increases the historical operator count by 1 and returns the new count
      */
-    function _updateTotalOperatorHistory(uint8 quorumNumber, uint32 numOperators) internal {
-        _totalOperatorsHistory[quorumNumber].push(QuorumUpdate({
-            numOperators: numOperators,
-            fromBlockNumber: uint32(block.number)
-        }));
+    function _increaseOperatorCount(uint8 quorumNumber) internal returns (uint32) {
+        QuorumUpdate storage lastUpdate = _latestQuorumUpdate(quorumNumber);
```
ChaoticWalrus, line 132: seems unrelated to this PR(?) but I think calling this struct 'QuorumUpdate' is super vague. It should probably be an 'OperatorCountUpdate' or something similar

src/IndexRegistry.sol
```diff
@@ -117,121 +126,170 @@ contract IndexRegistry is IndexRegistryStorage {
...
+        _updateOperatorCountHistory(quorumNumber, lastUpdate, newOperatorCount);
+
+        // If this is the first time we're using this index, push its first update
+        // This maintains an invariant: existing indices have nonzero history
+        if (_indexHistory[quorumNumber][newOperatorCount - 1].length == 0) {
+            _indexHistory[quorumNumber][newOperatorCount - 1].push(OperatorUpdate({
+                operatorId: OPERATOR_DOES_NOT_EXIST_ID,
+                fromBlockNumber: uint32(block.number)
+            }));
+        }
```
ChaoticWalrus, line 144: seems somewhat odd to push this here with the `OPERATOR_DOES_NOT_EXIST_ID` id?
I'll keep looking and try to circle back to this though

ChaoticWalrus: Ah, I see, it's necessary so that the history has nonzero length when we then call `_assignOperatorToIndex` immediately afterward.
Works for me.
I guess the property here to ensure we're using these consistently is (at minimum)
```
if (_indexHistory[quorumNumber][newOperatorCount - 1].length == 1) {
_indexHistory[quorumNumber][newOperatorCount - 1][0].operatorId != 0;
}
```
i.e. the end result should always be that the first update *ends up* with a nonzero `operatorId`, even though we set it to zero here

src/IndexRegistry.sol
```diff
@@ -117,121 +126,170 @@ contract IndexRegistry is IndexRegistryStorage {
...
-     * @param quorumNumber is the number of the quorum to update
-     * @param numOperators is the number of operators in the quorum
+     * @notice Increases the historical operator count by 1 and returns the new count
      */
-    function _updateTotalOperatorHistory(uint8 quorumNumber, uint32 numOperators) internal {
-        _totalOperatorsHistory[quorumNumber].push(QuorumUpdate({
-            numOperators: numOperators,
-            fromBlockNumber: uint32(block.number)
-        }));
+    function _increaseOperatorCount(uint8 quorumNumber) internal returns (uint32) {
```
ChaoticWalrus, line 131: feels like maybe a bit many / lot of layering of internal functions?
e.g. `_increaseOperatorCount` is only used in `registerOperator` and `_decreaseOperatorCount` is only used in `deregisterOperator`
not feeling super strongly opinionated about this, but I think this kind of structure can make it a little harder to follow quite what's happening in the logic

src/IndexRegistry.sol
```diff
@@ -117,121 +126,170 @@ contract IndexRegistry is IndexRegistryStorage {
...
-        // if `blockNumber` is from before the `quorumNumber` existed, return `0`
-        if (blockNumber < _totalOperatorsHistory[quorumNumber][0].fromBlockNumber) {
-            return 0;
-        }
+        uint256 historyLength = _operatorCountHistory[quorumNumber].length;
+        require(historyLength != 0, "IndexRegistry._operatorCountAtBlockNumber: quorum does not exist");
+        require(
+            blockNumber >= _operatorCountHistory[quorumNumber][0].fromBlockNumber, 
+            "IndexRegistry._operatorCountAtBlockNumber: quorum did not exist at given block number"
+        );
```
ChaoticWalrus, line 260: is this not redundant with the revert block at the end of the function? if we delete this then we should also delete the `// Shouldn't be able to reach this point` comment though

src/IndexRegistry.sol
```diff
@@ -244,65 +302,48 @@ contract IndexRegistry is IndexRegistryStorage {
                                  VIEW FUNCTIONS
     *******************************************************************************/
 
-    /// @notice Returns the _indexToOperatorIdHistory entry for the specified `operatorIndex` and `quorumNumber` at the specified `index`
-    function getOperatorIndexUpdateOfIndexForQuorumAtIndex(uint32 operatorIndex, uint8 quorumNumber, uint32 index) external view returns (OperatorUpdate memory) {
-        return _indexToOperatorIdHistory[quorumNumber][operatorIndex][index];
+    /// @notice Returns the _indexHistory entry for the specified `operatorIndex` and `quorumNumber` at the specified `index`
+    function getOperatorUpdateAtIndex(uint32 operatorIndex, uint8 quorumNumber, uint32 index) external view returns (OperatorUpdate memory) {
+        return _indexHistory[quorumNumber][operatorIndex][index];
```
ChaoticWalrus, line 307: feels weird that these orderings are different (function inputs vs. mapping structure)

src/IndexRegistry.sol
```diff
@@ -244,65 +302,48 @@ contract IndexRegistry is IndexRegistryStorage {
...
-            quorumOperatorList[i] = _getOperatorIdAtIndexForQuorumAtBlockNumber(uint32(i), quorumNumber, blockNumber);
+        uint32 operatorCount = _operatorCountAtBlockNumber(quorumNumber, blockNumber);
+        bytes32[] memory operatorList = new bytes32[](operatorCount);
+        for (uint256 i = 0; i < operatorCount; i++) {
+            operatorList[i] = _operatorIdForIndexAtBlockNumber(uint32(i), quorumNumber, blockNumber);
             require(
-                quorumOperatorList[i] != OPERATOR_DOES_NOT_EXIST_ID, 
-                "IndexRegistry.getOperatorListForQuorumAtBlockNumber: operator does not exist at the given block number"
+                operatorList[i] != OPERATOR_DOES_NOT_EXIST_ID, 
+                "IndexRegistry.getOperatorListAtBlockNumber: operator does not exist at the given block number"
```
ChaoticWalrus, line 338: this is effectively an assert here, right?
seems like by design there should be no path that ever allows hitting this require statement. If it's triggered, then at least one system property has been broken?

ChaoticWalrus (commented):
IndexRegistry looking good, really liking the changes to it overall; I added some comments but I don't think anything is big, just nits or discussion points. Will be proceeding with remainder of review.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -209,10 +195,10 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
         operatorIdsToSwap[0] = pubkey.hashG1Point();
 
         // verify the churnApprover's signature
-        _verifyChurnApproverSignatureOnOperatorChurnApproval({
+        _verifyChurnApproverSignature({
             registeringOperatorId: operatorIdsToSwap[0],
             operatorKickParams: operatorKickParams,
-            signatureWithSaltAndExpiry: signatureWithSaltAndExpiry
+            churnApproverSignature: churnApproverSignature
```
ChaoticWalrus, line 201: should we maybe do this earlier in this function?
seems like a fairly late revert condition, but I guess those are somewhat difficult to avoid in this design overall.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -57,12 +57,12 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
-    /// @notice the mapping from operator's operatorId to the updates of the bitmap of quorums they are registered for
-    mapping(bytes32 => QuorumBitmapUpdate[]) internal _operatorIdToQuorumBitmapHistory;
-    /// @notice the mapping from operator's address to the operator struct
-    mapping(address => Operator) internal _operators;
+    /// @notice maps quorum number => operator cap and kick params
+    mapping(uint8 => OperatorSetParam) internal _quorumParams;
+    /// @notice maps operator id => historical quorums they registered for
+    mapping(bytes32 => QuorumBitmapUpdate[]) internal _operatorBitmapHistory;
+    /// @notice maps operator address => operator id and status
+    mapping(address => Operator) internal _operatorInfo;
```
ChaoticWalrus, line 65: should we perhaps name the struct 'OperatorInfo' as well?

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -408,26 +374,26 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
             // new stored quorumBitmap is the previous quorumBitmap or'd with the new quorumBitmap to register for
             quorumBitmap |= prevQuorumBitmap;
 
-            _operatorIdToQuorumBitmapHistory[operatorId][historyLength - 1].nextUpdateBlockNumber = uint32(block.number);
+            _operatorBitmapHistory[operatorId][historyLength - 1].nextUpdateBlockNumber = uint32(block.number);
         }
 
         // set the operatorId to quorum bitmap history
-        _operatorIdToQuorumBitmapHistory[operatorId].push(QuorumBitmapUpdate({
+        _operatorBitmapHistory[operatorId].push(QuorumBitmapUpdate({
```
ChaoticWalrus, line 388: here we are still pushing an update, as opposed to modifying the previous update (if the block number is the same).
seems fine to implement later, just pointing this out

ChaoticWalrus: I see TODOs elsewhere so only called out this location

ChaoticWalrus (approved):
Changes LGTM.
Feels fine to merge this as-is; I looked through everything and don't have any substantive issues, mostly just suggestions for further changes.

src/IndexRegistry.sol
wadealexc: yep. i'm not super happy with the way i added this to the increase count method, but couldn't find a different way that kept things super simple. ill think about it this morning

src/IndexRegistry.sol
wadealexc: good call out, thanks. i was feeling similarly - but i think the last implementation was worse - since the layering was there, but a lot of helper functions were similarly-named and served similar functions, or called each other, etc.

i'm gonna keep this as is while i merge some of these PRs, but we should revisit this again and see if it's worth further changes.

src/IndexRegistry.sol
wadealexc: Good point, thanks. I removed this initial check and the final check's comment.

src/IndexRegistry.sol
wadealexc: Agreed. I'm not a fan of `operatorIndex` naming either. IMO the worst part of the `IndexRegistry` is trying to make the implicit structure of the "operator list at block number X" pop out from the code.

Part of the confusion is calling things an "index" sometimes, vs the "operator count" at others. I'm not sure how to marry these two concepts. If you have any ideas I'd love to hear them.

I'll change the param order :)

src/IndexRegistry.sol
wadealexc: 1. The lookup returns this value if the index was never used at the blocknumber.
2. When we remove an operator, the final entry in the index history ends up being this value

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: yes :)

i'm doing this in a future PR

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: +1. Will make this change in a future pr.

src/IndexRegistry.sol
wadealexc: see [231bbe3](https://github.com/Layr-Labs/eigenlayer-middleware/pull/43/commits/231bbe33b677545dbb092033f0be21fdfc83d501)

src/IndexRegistry.sol
wadealexc: see [231bbe3](https://github.com/Layr-Labs/eigenlayer-middleware/pull/43/commits/231bbe33b677545dbb092033f0be21fdfc83d501)

src/IndexRegistry.sol
wadealexc: i'm going to keep this for now, but we can revisit when the PRs are merged :)

PR #47, opened 2023-11-01
docs: track deployment

stevennevins:


ChaoticWalrus (approved):
Seems like a good addition to the README.

I took a quick look and the contracts aren't verified on Etherscan. We definitely want to make sure we do that on the final deployment, so if we intend to use these going forward we should probably do that? At the very least, making sure we note the commit used for their deployment is important for this.

stevennevins:
Yeah, I wanted to get what we had tracked these are from the most recent EigenDA deployment. I’ll deploy and have the output tracked in git and update the readme to link to the commit associated like we do on the main repo

ChaoticWalrus:
> Yeah, I wanted to get what we had tracked these are from the most recent EigenDA deployment. I’ll deploy and have the output tracked in git and update the readme to link to the commit associated like we do on the main repo

sounds like a good plan going forward, thanks!

PR #48, opened 2023-11-01
chore: remove unused functions from migration

stevennevins:
Remnant from before middleware was removed from core


ChaoticWalrus (approved):
Looks fine to me; I recall us agreeing on these changes previously.
I guess just make sure CI passes before merging?

PR #50, opened 2023-11-02
Feat: updateStakesAllOperators() for StakeRegistry pull-updates

8sunyuan:
Description: 
The idea is to add a method to enforce updates of all registered operators across all quorums. StakeRegistry now has `updateStakesAllOperators()` function which is the same as `updateStakes` but keeps track of the previous operatorId to ensure no duplicates and the number of updated operators matches `IndexRegistry.totalOperatorsForQuorum()`. 

Changes:
StakeRegistry.sol: added `updateStakesAllOperators` and now passing in IndexRegistry to the constructor
IStakeRegistry.sol: added `updateStakesAllOperators`

### DIFF
--- a/src/StakeRegistry.sol
+++ b/src/StakeRegistry.sol
@@ -5,6 +5,7 @@ import "eigenlayer-contracts/src/contracts/libraries/BitmapUtils.sol";
 import "src/interfaces/IServiceManager.sol";
 import "src/interfaces/IStakeRegistry.sol";
 import "src/interfaces/IRegistryCoordinator.sol";
+import "src/interfaces/IIndexRegistry.sol";
 import "src/StakeRegistryStorage.sol";
 import {VoteWeigherBase} from "src/VoteWeigherBase.sol";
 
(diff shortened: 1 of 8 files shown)


src/StakeRegistry.sol
```diff
@@ -124,6 +126,72 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+        // if they are, get their new weight and update their individual stake history and the
+        // quorum's total stake history accordingly
+        for (uint8 quorumNumber = 0; quorumNumber < quorumCount; ) {
+            OperatorStakeUpdate memory totalStakeUpdate;
+            uint256 numOperatorsUpdatedInQuorum = 0;
+            bytes32 prevOperatorId = bytes32(0);
+            // for each operator
+            for (uint i = 0; i < operators.length; ) {
+                bytes32 operatorId = registryCoordinator.getOperatorId(operators[i]);
+                uint192 quorumBitmap = registryCoordinator.getCurrentQuorumBitmapByOperatorId(operatorId);
```
gpsanant, line 149: this will get the bitmap and operator id for operators a bunch more times than it needs too... what if we just create a `uint32[] memory numOperatorsUpdatedInQuorum` and we loop thru each operator

8sunyuan: Agreed, maybe something like `bytes32[][] operatorIdsPerQuorum` as the input params? An array of operatorIds per quorum.
For each quorum, do a require check for `indexRegistry.totalOperatorsForQuorum(quorumNumber)` and the operatorIds array for that quorum. For each operatorId, check that the quorum number is in the corresponding bitmap.

PR #51, opened 2023-11-02
refactor: pubkey compendium stores pubkeys

wadealexc:
Builds on PR #28 and #43; merge those first! It looks like I'm merging 20 commits, but those are actually commits from the other PRs :) Only the final 2 commits in this PR are relevant.

Functional changes in this PR:
* adds a new mapping to `BLSPublicKeyCompendium` : `operatorToPubkey`, which maps operator addresses to `G1Point` 
* adds a method `BLSPublicKeyCompendium.getRegisteredPubkey(address)` which performs a pubkey lookup along with validation that the pubkey exists/isn't the zero pubkey. 
  * Validation logic here was previously performed in the `BLSPubkeyRegistry`
* Removes the "pubkey" parameter from various functions, like in the `BLSPubkeyRegistry` and `BLSRegistryCoordinatorWithIndices`. The pubkey registry is able to look up pubkeys with just an operator address, now.

`BLSPubkeyRegistry` state variable naming changes (these better match the rest of the contracts):
* `quorumApk` -> `currentApk`
* `quorumApkUpdates` -> `apkHistory` 

`BLSPubkeyRegistry` view function naming changes:
* `getApkIndicesForQuorumsAtBlockNumber` -> `getApkIndicesAtBlockNumber`
* `getApkForQuorum` -> `getApk`
* `getApkUpdateForQuorumByIndex` -> `getApkUpdateAtIndex`
* `getApkHashForQuorumAtBlockNumberFromIndex` -> `getApkHashAtBlockNumberAndIndex`
* `getQuorumApkHistoryLength` -> `getApkHistoryLength`

**Recommended review order:**
1. Check out the `BLSPublicKeyCompendium` changes in `b883fd0` to see what new state we're tracking and get some context for the other changes in this commit.
2. Review the rest of the commit, noting all the places we're no longer passing in pubkeys.
3. Skim the final commit - this is solely naming-related changes (listed above)

### DIFF
--- a/src/BLSPublicKeyCompendium.sol
+++ b/src/BLSPublicKeyCompendium.sol
@@ -12,10 +12,15 @@ import "eigenlayer-contracts/src/contracts/libraries/BN254.sol";
 contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
     using BN254 for BN254.G1Point;
 
-    /// @notice mapping from operator address to pubkey hash
+    /// @notice the hash of the zero pubkey aka BN254.G1Point(0,0)
+    bytes32 internal constant ZERO_PK_HASH = hex"ad3228b676f7d3cd4284a5443f17f1962b36e491b30a40b2405849e597ba5fb5";
+
+    /// @notice maps operator address to pubkey hash
     mapping(address => bytes32) public operatorToPubkeyHash;
-    /// @notice mapping from pubkey hash to operator address
+    /// @notice maps pubkey hash to operator address
     mapping(bytes32 => address) public pubkeyHashToOperator;
+    /// @notice maps operator address to pubkeyG1
+    mapping(address => BN254.G1Point) public operatorToPubkey;
 
     /*******************************************************************************
                             EXTERNAL FUNCTIONS 
@@ -33,6 +38,9 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
         BN254.G2Point memory pubkeyG2
     ) external {
         bytes32 pubkeyHash = BN254.hashG1Point(pubkeyG1);
+        require(
+            pubkeyHash != ZERO_PK_HASH, "BLSPublicKeyCompendium.registerBLSPublicKey: cannot register zero pubkey"
+        );
         require(
             operatorToPubkeyHash[msg.sender] == bytes32(0),
             "BLSPublicKeyCompendium.registerBLSPublicKey: operator already registered pubkey"
@@ -65,6 +73,7 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
             pubkeyG2
         ), "BLSPublicKeyCompendium.registerBLSPublicKey: either the G1 signature is wrong, or G1 and G2 private key do not match");
 
+        operatorToPubkey[msg.sender] = pubkeyG1;
         operatorToPubkeyHash[msg.sender] = pubkeyHash;
         pubkeyHashToOperator[pubkeyHash] = msg.sender;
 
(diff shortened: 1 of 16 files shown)


src/BLSPublicKeyCompendium.sol
```diff
@@ -75,6 +84,24 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
...
+    function getRegisteredPubkey(address operator) public view returns (BN254.G1Point memory) {
+        BN254.G1Point memory pubkey = operatorToPubkey[operator];
+        bytes32 pubkeyHash = operatorToPubkeyHash[operator];
+
+        require(
+            pubkeyHash != bytes32(0) && BN254.hashG1Point(pubkey) == pubkeyHash, 
+            "BLSPublicKeyCompendium.getRegisteredPubkey: operator is not registered"
+        );
+
+        require(pubkeyHash != ZERO_PK_HASH, "BLSPublicKeyCompendium.getRegisteredPubkey: invalid pubkey");
```
gpsanant, line 100: don't need this

wadealexc: Because it's checked implicitly above?

src/BLSPublicKeyCompendium.sol
```diff
@@ -75,6 +84,24 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
...
+     * @dev Reverts if the operator has not registered a valid pubkey
+     */
+    function getRegisteredPubkey(address operator) public view returns (BN254.G1Point memory) {
+        BN254.G1Point memory pubkey = operatorToPubkey[operator];
+        bytes32 pubkeyHash = operatorToPubkeyHash[operator];
+
+        require(
+            pubkeyHash != bytes32(0) && BN254.hashG1Point(pubkey) == pubkeyHash, 
+            "BLSPublicKeyCompendium.getRegisteredPubkey: operator is not registered"
+        );
```
ChaoticWalrus, line 98: the second part of this check feels like overkill / not the right place to be enforcing this?
we are already enforcing that `BN254.hashG1Point(operatorToPubkey[operator]) == operatorToPubkeyHash[operator]`
I'm not opposed to some redundant checks in general, but I think a specific benefit of this storage change could be _not_ needing to repeatedly do the `hashG1Point` operation, which this seems to be doing unnecessarily

src/BLSPublicKeyCompendium.sol
```diff
@@ -33,6 +38,9 @@ contract BLSPublicKeyCompendium is IBLSPublicKeyCompendium {
         BN254.G2Point memory pubkeyG2
     ) external {
         bytes32 pubkeyHash = BN254.hashG1Point(pubkeyG1);
+        require(
+            pubkeyHash != ZERO_PK_HASH, "BLSPublicKeyCompendium.registerBLSPublicKey: cannot register zero pubkey"
+        );
```
ChaoticWalrus, line 43: seems odd we weren't doing this before. @gpsanant is there something else preventing someone from registering the zero key? Lack of knowledge of the corresponding private key?
guess we could check that the G1 point doesn't equal the zero point, but this is equivalent 🤷

ChaoticWalrus (approved):
Loving this PR, thank you!
I have been in favor of this storage change for a while, and this PR really helps make the simplifications it enables clear.
Renaming changes LGTM as well.
I noted one place where we are doing a redundant (and as far as I understand potentially expensive?) check, but we can debate that later / I wouldn't consider resolving the discussion to be blocking for this PR, personally.

src/BLSPublicKeyCompendium.sol
samlaf: Agree that consistency checks between the mapping should be enforced at write time, not read time. Is there a way they can diverge?

src/BLSPublicKeyCompendium.sol
samlaf: You’re already not allowing zero pubkey to register right?  So it would be impossible for this check to ever fail I think?

src/BLSPublicKeyCompendium.sol
samlaf: I'm also curious about this. private key for zero point is just 0.
Seems like one could register by using pkG2=sigG1=0 ? But don't understand pairings enough to know whether `e(sigG1,G2)=e(h(m),pkG2)` since I think (?) `e(0,anything) = e(anything,0) = 0` (it's a multilinear map after all)

src/BLSPublicKeyCompendium.sol
wadealexc: see [bb4011e](https://github.com/Layr-Labs/eigenlayer-middleware/pull/51/commits/bb4011ede8aa3e0ee29133d38487530f12806056)

src/BLSPublicKeyCompendium.sol
wadealexc: see [bb4011e](https://github.com/Layr-Labs/eigenlayer-middleware/pull/51/commits/bb4011ede8aa3e0ee29133d38487530f12806056)

src/BLSPublicKeyCompendium.sol
gpsanant: this isn't really a substantive check, it's not needed, it doesn't prevent any attacks

ChaoticWalrus: so it might not prevent an attack per se, but it would prevent weird behavior, right?
the fact that the apk doesn't change when zero is added to or subtracted from it feels like it could lead to some strange effects. For example, any signature would appear as if you both did and didn't sign it, depending on the interpretation in the context of the code.
I get that it's like using a key that everyone else knows, but it also has this special aspect. Seems safer to avoid the possibility of it happening.

wadealexc: > this isn't really a substantive check, it's not needed, it doesn't prevent any attacks

i think it's substantive now, since nowhere else checks this. i got rid of the checks against the zero hash in the pk registry in favor of checking it here

PR #53, opened 2023-11-06
fix: registry coordinator churn now uses correct kick index

wadealexc:
Needs a regression test; will open an issue.


0x0aa0 (approved):

ChaoticWalrus (approved):
good catch!

PR #55, opened 2023-11-07
chore: move BitmapUtils and BN254 libraries to middleware

wadealexc:
Moves these libraries to middleware, since they're only used here

Depends on first merging core PR: https://github.com/Layr-Labs/eigenlayer-contracts/pull/319

Once that's merged I'll push a final commit to update submodules, so please go review the above PR first!

Edit: Ready for merge!


ChaoticWalrus (approved):
gud change

PR #56, opened 2023-11-07
chore: merge master into m2-mainnet

wadealexc:


wadealexc:
force-merging because i cant fix the old commits not adhereing to the linter

samlaf:
@wadealexc can we rebase downstream and not merge? Branch history becomes impossible to understand and reason when merging downstream.

PR #57, opened 2023-11-07
refactor: move updateStakes to registry coordinator

wadealexc:


src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -337,109 +332,183 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
         bytes calldata quorumNumbers,
-        string memory socket
-    ) internal virtual returns(uint32[] memory, bytes32) {        
-        // Create and validate bitmap from quorumNumbers
-        uint256 quorumBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers);
-        require(quorumBitmap <= MAX_QUORUM_BITMAP, "BLSRegistryCoordinatorWithIndices._registerOperator: bitmap exceeds max bitmap size");
-        require(quorumBitmap != 0, "BLSRegistryCoordinatorWithIndices._registerOperator: bitmap cannot be 0");
-        
+        string memory socket,
+        bool performChurn,
```
ChaoticWalrus, line 344: seems like an unnecessary / duplicate parameter?

ChaoticWalrus: ahh, I guess it enables the clearer 'churn required for overfilled quorum' error below, rather than simply hitting an index out-of-bounds error. seems OK then

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -337,109 +332,183 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+        require(kickParams.quorumNumber == quorumNumber, "BLSRegistryCoordinatorWithIndices._validateChurn: quorumNumber not the same as signed");
+
+        // Get the target operator's stake and check that it is below the kick thresholds
+        uint96 operatorToKickStake = stakeRegistry.getCurrentOperatorStakeForQuorum(idToKick, quorumNumber);
+        require(
+            newOperatorStake > _individualKickThreshold(operatorToKickStake, setParams),
+            "BLSRegistryCoordinatorWithIndices._validateChurn: incoming operator has insufficient stake for churn"
+        );
+        require(
+            operatorToKickStake < _totalKickThreshold(totalQuorumStake, setParams),
```
ChaoticWalrus, line 447: extremely minor nit, but should these maybe be non-strict inequalities?

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -337,109 +332,183 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+    function _individualKickThreshold(uint96 operatorStake, OperatorSetParam memory setParams) internal pure returns (uint96) {
+        return operatorStake * setParams.kickBIPsOfOperatorStake / BIPS_DENOMINATOR;
+    }
+
+    /**
+     * @notice Returns the total stake threshold required for an operator to remain in a quorum.
+     * The operator must have at least the returned stake amount to keep their position.
+     */
+    function _totalKickThreshold(uint96 totalStake, OperatorSetParam memory setParams) internal pure returns (uint96) {
+        return totalStake * setParams.kickBIPsOfTotalStake / BIPS_DENOMINATOR;
```
ChaoticWalrus, line 511: might want public getters that use these + read from state?

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: Yes, also this most closely mimics the current implementation, which performs a loop over each quorum after registration and checks if the new operator count is over the max.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -486,6 +555,53 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+            } else {
+                lastUpdate.nextUpdateBlockNumber = uint32(block.number);
+                _operatorBitmapHistory[operatorId].push(QuorumBitmapUpdate({
+                    updateBlockNumber: uint32(block.number),
+                    nextUpdateBlockNumber: 0,
+                    quorumBitmap: newBitmap
+                }));
+            }
+        }
+    }
```
ChaoticWalrus, line 592: one somewhat subtle change in behavior here:
when an operator totally deregisters, previously their last bitmap entry would look like:
```
updateBlockNumber: X
nextUpdateBlockNumber: deregistrationBlock
quorumBitmap: lastNonzeroBitmap
```
but now we will push another entry that looks like:
```
updateBlockNumber: deregistrationBlock
nextUpdateBlockNumber: 0
quorumBitmap: 0
```
I believe the main "property" change here is that before we (should have) never pushed an update with an empty bitmap, and now we definitely will be.
Seems fine + I believe all behavior is consistent.

wadealexc: Yes, exactly - I believe that's in the changelog! The motivation was partially to be consistent with the "latest historical entry" as the operator's current status, because previously it wasn't.

Also this introduces some nice invariants around OperatorStatus that we can write tests for - namely, that if an operator is REGISTERED, they have a nonzero bitmap, and if they're NEVER_REGISTERED or DEREGISTERED, they have a zero bitmap.

src/libraries/BitmapUtils.sol
```diff
@@ -91,6 +91,25 @@ library BitmapUtils {
...
+     * @dev Reverts if bitmap contains a bit greater than or equal to `maxSetBit`
+     */
+    function orderedBytesArrayToBitmap(bytes memory orderedBytesArray, uint8 maxSetBit) internal pure returns (uint256) {
+        uint256 bitmap = orderedBytesArrayToBitmap(orderedBytesArray);
+
+        if (bitmap != 0) {
+            require(
+                uint8(orderedBytesArray[orderedBytesArray.length - 1]) < maxSetBit, 
+                "BitmapUtils.orderedBytesArrayToBitmap: bitmap exceeds max value"
+            );
```
ChaoticWalrus, line 107: this isn't the right check
it should be checking that `bitmap < (2>> maxSetBit)`
with this check as-is, you could have a length 1 input that just set a single, too-high bit number

ChaoticWalrus: I guess really should be `bitmap < (2>> (maxSetBit + 1))` if you're saying that the max bit is acceptable.
only edge case is if `maxSetBit` is 255 then

wadealexc: hmm... so originally i had this mirror the validation we were doing that `bitmap <= MAX_QUORUM_BITMAP`

But then i realized what we actually wanted to check for was that the highest set bit was less than the current quorum count

wadealexc: > with this check as-is, you could have a length 1 input that just set a single, too-high bit number

not sure i follow, wouldn't this be caught?

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -337,109 +332,183 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                _validateChurn({
+                    quorumNumber: quorumNumber,
+                    totalQuorumStake: totalStakes[i],
+                    newOperator: operatorToRegister,
+                    newOperatorStake: registeredStakes[i],
+                    kickParams: operatorKickParams[i],
+                    setParams: operatorSetParams
+                });
+
+                _deregisterOperator(operatorKickParams[i].operator, quorumNumbers[i:i+1]);
```
ChaoticWalrus, line 420: is `quorumNumbers[i:i+1]` the correct syntax, or should it just be `quorumNumbers[i]`?
I feel like this might be doing 2 quorums instead of 1 right now?

wadealexc: actually just borrowed this from the current implementation. it looks like the correct syntax to get a slice of length 1, but i'd hope we're testing it somewhere :D

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -337,109 +332,183 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+             * If we exceed the max operator count for this quorum, we need to replace an existing operator
+             * with a new one.
+             */
+            if (exceedsOperatorCapacity) {
+                require(performChurn, "BLSRegistryCoordinatorWithIndices._registerOperator: churn required for overfilled quorum");
+                
+                // Validate that `operatorToRegister` can replace the operator specified in the kick params
+                _validateChurn({
+                    quorumNumber: quorumNumber,
+                    totalQuorumStake: totalStakes[i],
```
ChaoticWalrus, line 413: this is now getting validated against the _new_ total stakes, when before I think it was the _old_ total stakes (please correct me if I'm wrong!).
I don't think either option is "incorrect" but we should make the intention clear

wadealexc: I think it's actually the same as before, glad you caught this though.

In both prev and cur implementation, we're updating the incoming operator's stake (and the quorum total stake) when the operator is registered.

We're then doing this validation against up-to-date info for the incoming operator and total, but potentially-old info the for operator getting churned.

ChaoticWalrus (changes requested):
I like these changes overall.
Feels good in terms of clarity, especially.
Noted a few places where behavior has notably changed; I think these are fine but it'd be good for us to specify the desired behavior / understand how properties change as a result of these modifications.
Think I've identified one or two bugs to be resolved prior to merging.

src/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus: ahhh OK. Makes sense; I guess I hadn't considered the two options and it just occurred to me then.

src/StakeRegistry.sol
8sunyuan, line 1: Don't know how to reference the line in large diff but for `_weightOfOperatorForQuorum` you have the return booleans mixed. Change to following

```
        // Return the weight, and `true` if the operator meets the quorum's minimum stake
        if (weight < minimumStakeForQuorum[quorumNumber]) {
            return (weight, false);
        } else {
            return (weight, true);
        }
```

src/interfaces/IStakeRegistry.sol
```diff
@@ -231,8 +236,15 @@ interface IStakeRegistry is IRegistry {
     function getCurrentTotalStakeForQuorum(uint8 quorumNumber) external view returns (uint96);
 
     /**
-     * @notice Used for updating information on deposits of nodes.
-     * @param operators are the addresses of the operators whose stake information is getting updated
+     * @notice Called by the registry coordinator to update an operator's stake for one
+     * or more quorums.
+     *
```
8sunyuan, line 241: Add @return natspec here. We should probably be more consistent on our natspec docs throughout but especially for interfaces where we don't see the implementation.

src/StakeRegistry.sol
wadealexc: Nice catch. I'll change it to work without the if/else syntax :) 

```solidity
bool hasMinimumStake = weight > minimumStakeForQuorum[quorumNumber];
return (weight, hasMinimumStake);
```

src/StakeRegistry.sol
```diff
@@ -187,14 +137,65 @@ contract StakeRegistry is StakeRegistryStorage {
...
+            require(_quorumExists(quorumNumber), "StakeRegistry.updateOperatorStake: quorum does not exist");
+
+            // Fetch the operator's current stake, applying weighting parameters and checking
+            // against the minimum stake requirements for the quorum.
+            (uint96 stakeWeight, bool hasMinimumStake) = _weightOfOperatorForQuorum(quorumNumber, operator);
+
+            // If the operator no longer meets the minimum stake, set their stake to zero and mark them for removal
+            if (!hasMinimumStake) {
+                stakeWeight = 0;
+                quorumsToRemove |= quorumNumber;
```
8sunyuan, line 173: Rather than using a bitmap for `quorumsToRemove` I think it makes more sense to use a bytes array instead here. This is only called in `updateOperators()` in the registryCoordinator which converts it back to a bytes array anyways for the deregistering

wadealexc: I hear you, but the additional syntax required to create that dynamic array in `updateOperatorStake` is super annoying.

And the end result is probably similar in gas cost, if i had to guess.

src/StakeRegistry.sol
wadealexc: See [bc7ed88](https://github.com/Layr-Labs/eigenlayer-middleware/pull/57/commits/bc7ed88dc83db35525a2794f5d5495eb8dea99d5)

src/interfaces/IStakeRegistry.sol
wadealexc: ... and [35247f8](https://github.com/Layr-Labs/eigenlayer-middleware/pull/57/commits/35247f8e1458ae2a50f32192141c2ec76f204ec1) because me dumb

src/libraries/BitmapUtils.sol
ChaoticWalrus: adding here: I read this code wrong, that's my bad 🤦  This looks correct.

src/StakeRegistry.sol
wadealexc: and [15c18a8](https://github.com/Layr-Labs/eigenlayer-middleware/pull/57/commits/15c18a8253f796a27cd3d86edf041fee1f9fd9c0)

because the operator should also have minimum if equal to the minimum for the quorum!

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: I split the churn function back out of this method and into the top-level method. Needed to in order to fix the stack too deep issue. Anyway, I'm not too mad. I think it's a bit easier to understand now.

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: Question for @gpsanant 

I feel like maybe the top one makes sense as a non-strict inequality, but the bottom doesnt?

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: I'm open to it if it's needed.

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: @gpsanant would appreciate your confirming that this is correct.

src/StakeRegistry.sol
8sunyuan: Sounds good, yeah probably not huge gas savings

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -259,6 +252,43 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+            }
+
+            uint192 currentBitmap = _currentOperatorBitmap(operatorId);
+            bytes memory currentQuorums = BitmapUtils.bitmapToBytesArray(currentBitmap);
+
+            /**
+             * Update the operator's stake for their active quorums. The stakeRegistry returns a bitmap
+             * of quorums where the operator no longer meets the minimum stake, and should be deregistered.
+             */
+            (uint192 quorumsToRemove) = stakeRegistry.updateOperatorStake(operator, operatorId, currentQuorums);
```
gpsanant, line 281: why parentheses

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -143,35 +146,46 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+     * operator capacity, this method will fail.
+     * @param quorumNumbers is an ordered byte array containing the quorum numbers being registered for
      */
     function registerOperator(
         bytes calldata quorumNumbers,
         string calldata socket
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
-        (uint32[] memory numOperatorsPerQuorum, ) = _registerOperator({
-            operator: msg.sender, 
+        bytes32 operatorId = blsPubkeyRegistry.getOperatorId(msg.sender);
```
gpsanant, line 157: why do we need this idToRegister stuff?

src/BLSRegistryCoordinatorWithIndices.sol
gpsanant: lets keep it strict for now, that's what's in the offchain code

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: oh yeah, that's from an earlier change. ill fix the naming, good call out

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: from an earlier change i think? anyway, ill remove em.

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: Done in [23eb1a3](https://github.com/Layr-Labs/eigenlayer-middleware/pull/57/commits/23eb1a3e7cf4f2e4f81af138d307a31355e56cb5)

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: Done in [23eb1a3](https://github.com/Layr-Labs/eigenlayer-middleware/pull/57/commits/23eb1a3e7cf4f2e4f81af138d307a31355e56cb5)

PR #62, opened 2023-11-09
Feat(M2-Mainnet): StakeRegistry pull updates per quorum

8sunyuan:
Description:
Due to gas costs for updating all operators in all quorums, we will implement pull updates on a per quorum basis.
More specifically we can first call `updateOperatorsPerQuorum()` for quorum0 anticipating a larger number of operators and strategies and then `updateOperatorsPerQuorum()` for the remaining quorums. 
Note: With this in place for m2-mainnet, we will also remove all push updates in the DelegationManager

Changes:
- `RegistryCoordinator.updateOperatorsPerQuorum()` which updates stakes for all operators for the specified quorums. Operators must be passed in based on ascending operatorId per quorum to ensure uniqueness for length checks.
- The StakeRegistry now keeps track of the last update timestamps for each quorum in the mapping `quorumUpdateTimestamp`.
This is only updated if all operators are updated at once from calling `updateOperatorsPerQuorum()` which calls `StakeRegistry.updateQuorumTimestamp(uint8 quorumNumber)`
- `BLSSignatureChecker.checkSignatures` now ensures that for each quorum, the quorumUpdateTimestamp isn't staler than a week.


src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,66 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+    /**
+     * @notice Updates the stakes of all operators for each of the specified quorums in the StakeRegistry. Each quorum also
+     * has their StakeRegistry.quorumTimestamp updated. which is meant to keep track of when operators were last all updated at once. 
+     * @param operatorsPerQuorum is an array of arrays of operators to update for each quorum. Note that each nested array
+     * of operators must be sorted in ascending operatorId order to ensure that all operators in the quorum are updated
+     * @param quorumNumbers is an array of quorum numbers to update
+     * @dev This method is used to update the stakes of all operators in a quorum at once, rather than individually
+     * 
+     */
+    function updateOperatorsPerQuorum(address[][] calldata operatorsPerQuorum, bytes calldata quorumNumbers) external onlyWhenNotPaused(PAUSED_UPDATE_OPERATOR) {
```
gpsanant, line 301: prefer `ForQuorum`

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,66 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                address operator = currQuorumOperators[j];
+                bytes32 operatorId = getOperatorId(operator);
+                uint192 currentBitmap = _currentOperatorBitmap(operatorId);
+
+                /**
+                 * Assuming the operatorId cannot be 0, for the first operatorId, prevOperatorId will be 0
+                 * Require check is to prevent duplicate operators and that all quorum operators are updated
+                 */
+                require(
+                    operatorId > prevOperatorId,
```
gpsanant, line 326: this is kind of confusing for the caller, could we make sure the operatorAddresses are in order instead?

src/StakeRegistry.sol
```diff
@@ -190,6 +190,16 @@ contract StakeRegistry is StakeRegistryStorage {
         return quorumsToRemove;
     }
 
+    /**
+     * @notice Called by the registry coordinator when all operators for a quorum have been updated
+     * all at once. Updates to the current block timestamp
+     * @param quorumNumber the specific quorum that has been updated
+     */
+    function updateQuorumTimestamp(uint8 quorumNumber) external onlyRegistryCoordinator {
```
gpsanant, line 198: why not keep this in the reg coor?

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,66 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                require(
+                    BitmapUtils.numberIsInBitmap(currentBitmap, quorumNumber),
+                    "BLSRegistryCoordinatorWithIndices.updateOperatorsPerQuorum: operator not in quorum"
+                );
+
+                /**
+                 * Update the operator's stake for their active quorums. The stakeRegistry returns a bitmap
+                 * of quorums where the operator no longer meets the minimum stake, and should be deregistered.
+                 */
+                uint192 quorumsToRemove = stakeRegistry.updateOperatorStake(operator, operatorId, BitmapUtils.bitmapToBytesArray(currentBitmap));
```
gpsanant, line 338: shouldn't we be updating for the given quorum number, not the entire bitmap?

8sunyuan: Ahhh good call, should just be the quorum number.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,66 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+        require(
+            operatorsPerQuorum.length == quorumNumbers.length,
+            "BLSRegistryCoordinatorWithIndices.updateOperatorsPerQuorum: input length mismatch"
+        );
+        for (uint256 i = 0; i < quorumNumbers.length; ++i) {
+            uint8 quorumNumber = uint8(quorumNumbers[i]);
+            address[] memory currQuorumOperators = operatorsPerQuorum[i];
+            require(
+                currQuorumOperators.length == indexRegistry.totalOperatorsForQuorum(quorumNumber),
+                "BLSRegistryCoordinatorWithIndices.updateOperatorsPerQuorum: number of updated operators does not match quorum total"
```
wadealexc, line 311: Why do we need to do all the operators at a time? `updateOperators` doesn't require anything similar - does this have a use here?

wadealexc: What if the operator count/update cost gets high enough that we need to break it up again?

src/StakeRegistry.sol
wadealexc: Agreed

8sunyuan: I figured the StakeRegistry should be keeping track of the timestamp mapping but we can remove the external call and keep it in the coordinator

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,66 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+            );
+            bytes32 prevOperatorId = bytes32(0);
+
+            // Update stakes for each operator in this quorum
+            for (uint256 j = 0; j < currQuorumOperators.length; ++j) {
+                address operator = currQuorumOperators[j];
+                bytes32 operatorId = getOperatorId(operator);
+                uint192 currentBitmap = _currentOperatorBitmap(operatorId);
```
wadealexc, line 319: Would prefer we do a similar pattern like the one in `updateOperators` that validates that the `operator` has `OperatorStatus.REGISTERED`

wadealexc: Cause currently this allows you to update without being registered (i think). `_currentOperatorBitmap` returns 0 if you're not registered.

wadealexc: ah i see there's a check that the number is in the bitmap. Regardless - I like the explicit check; makes it easier to surface and enforce the status invariants

src/StakeRegistry.sol
wadealexc: updating operators means keeping their operator status up-to-date, as well as their listings in each of the registries. for that reason, i think the timestamp is a coordinator concept

wadealexc:
Overall, fine with these changes - but I wonder if `updateOperators` and `updateOperatorsPerQuorum` can be refactored to use a common helper method. They're almost identical.

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: Changed https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/99e543c4449e67f25ff3f94eed5d36f4e8099379

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/99e543c4449e67f25ff3f94eed5d36f4e8099379

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/99e543c4449e67f25ff3f94eed5d36f4e8099379

wadealexc (approved):
left a few questions, nothing blocking except for updating tests!

would not recommend writing new tests unless you want to dive into middleware tests in general. if you do want that, DM me and ill get you up to speed

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: I thought about that too, but we have to ensure the operator in the array is registered for the corresponding quorum. Not just at least one quorum

wadealexc: Yeah - I'm not suggesting removing specific-quorum check, just explicitly making the REGISTERED check - just so it's explicit and clear. It shouldn't be an extra storage read, since you're doing the same SLOAD to get the operatorId

wadealexc: It's nbd if you feel it adds clutter.

Maybe a compromise could be to add a comment above the specific quorum check, that this enforces the REGISTERED status invariant?

ChaoticWalrus: IMO the explicit REGISTERED check would be clearest, but I think the comment would be fine. really good point about us already SLOAD'ing the same slot.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,65 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                require(
+                    BitmapUtils.numberIsInBitmap(currentBitmap, quorumNumber),
+                    "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: operator not in quorum"
+                );
+
+                /**
+                 * Update the operator's stake for their active quorums. The stakeRegistry returns a bitmap
+                 * of quorums where the operator no longer meets the minimum stake, and should be deregistered.
+                 */
+                uint192 quorumsToRemove = stakeRegistry.updateOperatorStake(operator, operatorId, BitmapUtils.bitmapToBytesArray(1<<quorumNumber));
```
ChaoticWalrus, line 337: The call to `BitmapUtils.bitmapToBytesArray` is ~fairly gas-expensive here, as it explicitly does 256 loop iterations.
This function should only be used in general when it's needed.
We should be able to do a byte slice here instead, like `quorumNumbers[i:i+1]` -- I can't remember where, but we are already using this in at least one other middleware contract.

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -289,6 +289,65 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+     * @dev This method is used to update the stakes of all operators in a quorum at once, rather than individually
+     * 
+     */
+    function updateOperatorsForQuorum(address[][] calldata operatorsPerQuorum, bytes calldata quorumNumbers) external onlyWhenNotPaused(PAUSED_UPDATE_OPERATOR) {
+        require(
+            operatorsPerQuorum.length == quorumNumbers.length,
+            "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: input length mismatch"
+        );
+        for (uint256 i = 0; i < quorumNumbers.length; ++i) {
+            uint8 quorumNumber = uint8(quorumNumbers[i]);
```
ChaoticWalrus, line 307: it doesn't seem strictly necessary, but we might want some kind of sanitization here on the `quorumNumbers` input?
the main thing that strikes me is that you could input an invalid quorum number? but I guess it'd effectively be a no-op

src/BLSSignatureChecker.sol
```diff
@@ -73,6 +73,10 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
         // loop through every quorumNumber and keep track of the apk
         BN254.G1Point memory apk = BN254.G1Point(0, 0);
         for (uint i = 0; i < quorumNumbers.length; i++) {
+            require(
+                stakeRegistry.quorumUpdateTimestamp(uint8(quorumNumbers[i])) + 1 weeks <= block.timestamp,
+                "BLSSignatureChecker.checkSignatures: StakeRegistry updates must be within a week"
+            );
```
ChaoticWalrus, line 79: we had discussed that there should likely be a switch for enabling/disabling this check.

this seems correct as a first pass, but we should probably figure out where the switch lives + access control.

the `1 weeks` here is also a "magic number" -- it would be better practice to name this constant, I think. It also seems possible we want this to be a variable rather than a constant, so might be good to bring that up in discussion with the AVS team?

ChaoticWalrus (approved):
Looking good!
Left some minor feedback in comments.

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: It’s this contract. Good call

src/BLSRegistryCoordinatorWithIndices.sol
wadealexc: Oh yeah, the other methods all have this check. _quorumExists

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: It's more efficient for checking in the BLSSignatureChecker if we just check the updated timestamp for the quorum itself rather than keeping track of the lastest updates for each operator in the apk. Hopefully the added gas complexity with `_deregisterOperator()` in the new refactor doesn't make it infeasible to update all the operators in a single quorum.

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/0556f9276126cebd4eecde019282670365a175db
Added a modifier for this now

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/0556f9276126cebd4eecde019282670365a175db Updated with bytes slicing here as well as for the _deregisterOperator call since quorumsToRemove will only be 0 or with the specific quorum bit toggled

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/0556f9276126cebd4eecde019282670365a175db
Added check similar to `updateOperators` pattern

src/StakeRegistry.sol
8sunyuan: Moved to reg coordinator https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/6e650e5550e57bd1d071cb42072808d2f5558b70

wadealexc:
I think using the modifier makes it hard to use this check in other places, so I did a quick refactor and moved it to a helper. Here's my take; you're welcome to squash merge into this PR if you like it: https://github.com/Layr-Labs/eigenlayer-middleware/pull/64

wadealexc:
I'm otherwise fine with these changes for now, but may want to revisit when we start writing tests for the reg coord. I can't help but feel that with this additional method, we can shift some stuff around a bit.

`updateOperators` and `updateOperatorsPerQuorum` feel similar enough that I want to say there's a helper method that can support both.

And `_deregisterOperator` is now frequently called with a slice of 1 byte. I wonder if that can be improved somehow.

8sunyuan:
> I'm otherwise fine with these changes for now, but may want to revisit when we start writing tests for the reg coord. I can't help but feel that with this additional method, we can shift some stuff around a bit.
> 
> `updateOperators` and `updateOperatorsPerQuorum` feel similar enough that I want to say there's a helper method that can support both.
> 
> And `_deregisterOperator` is now frequently called with a slice of 1 byte. I wonder if that can be improved somehow.

Good points on potential refactoring, before going forward with these changes I've got to first get some gas estimations again on the new functionality.

src/BLSSignatureChecker.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/a4212dac9f000a0e73f2b99be7a69daef3520453
Updated it now to constant. Will bring this up later in discussion

src/BLSSignatureChecker.sol
```diff
@@ -74,8 +78,8 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
         BN254.G1Point memory apk = BN254.G1Point(0, 0);
         for (uint i = 0; i < quorumNumbers.length; i++) {
             require(
-                stakeRegistry.quorumUpdateTimestamp(uint8(quorumNumbers[i])) + 1 weeks <= block.timestamp,
-                "BLSSignatureChecker.checkSignatures: StakeRegistry updates must be within a week"
+                registryCoordinator.quorumUpdateTimestamp(uint8(quorumNumbers[i])) + delegation.withdrawalDelayBlocks() <= block.timestamp,
```
gpsanant, line 81: this needs to be a block, not a timestamp to add blocks to it

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -561,6 +545,29 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+        OperatorInfo memory operatorInfo,
+        bytes memory quorumsToUpdate
+    ) internal {
+        if (operatorInfo.status != OperatorStatus.REGISTERED) {
+            return;
+        }
+        bytes32 operatorId = operatorInfo.operatorId;
+        uint192 quorumsToRemove = stakeRegistry.updateOperatorStake(operator, operatorId, quorumsToUpdate);
+
+        if (!quorumsToRemove.isEmpty()) {
```
gpsanant, line 563: cant we check quorumsToUpdate.length == 0

gpsanant, line 563: cant we check quorumsToUpdate.length == 0

8sunyuan: Some quorums might fall below minimum stake and are marked to be removed in the `quorumToRemove` bitmap from the `stakeRegistry.updateOperatorStake` call so we check the value it returns

src/BLSSignatureChecker.sol
8sunyuan: fixed https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/7db7e0a73720d6cf598a43e8eae52399ecf138fb although doesn't compile at the moment while lib/eigenlayer-contracts is pointed to master branch instead of m2-mainnet. I added the interface function here https://github.com/Layr-Labs/eigenlayer-contracts/pull/344

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -261,31 +264,69 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                        operator > prevOperatorAddress,
+                        "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: operators array must be sorted in ascending address order"
+                    );
+                }
+                _updateOperator(operator, operatorInfo, quorumNumbers[i:i+1]);
             }
+
+            // Update timestamp that all operators in quorum have been updated all at once
+            quorumUpdateBlocknumber[quorumNumber] = block.number;
+            emit QuorumBlocknumberUpdated(quorumNumber, block.number);
```
gpsanant, line 329: QuorumUpdateBlockNumberUpdated? lol idk

src/BLSSignatureChecker.sol
```diff
@@ -28,12 +28,22 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
     IStakeRegistry public immutable stakeRegistry;
     IBLSPubkeyRegistry public immutable blsPubkeyRegistry;
     IDelegationManager public immutable delegation;
+    IServiceManager public immutable serviceManager;
+    /// @notice If true, check that the signature timestamp is within the delegation withdrawalDelayBlocks window.
+    bool public isTimestampChecked;
```
gpsanant, line 33: prefer something like `isStakeUpdateChecked`

test/mocks/RegistryCoordinatorMock.sol
```diff
@@ -46,5 +46,5 @@ contract RegistryCoordinatorMock is IRegistryCoordinator {
 
     function deregisterOperator(bytes calldata quorumNumbers, bytes calldata) external {}
 
-    function quorumUpdateTimestamp(uint8 quorumNumber) external view returns (uint256) {}
+    function quorumUpdateBlocknumber(uint8 quorumNumber) external view returns (uint256) {}
```
gpsanant, line 49: BlockNumber

src/BLSSignatureChecker.sol
ChaoticWalrus: perhaps `timelyUpdatesRequired` or something?
maybe `staleStakesForbidden` ?

ChaoticWalrus: got G's opinion offline, feels like `staleStakesForbidden` is best -- it feels succinct-ish with the most clarity of meaning.
`isTimestampChecked` and `isStakeUpdateChecked` beg the question "check for what / checked how?", and `timelyUpdatesRequired` begs the question "what kind of update?"

ChaoticWalrus: could be good to workshop /  ask a bit more broadly for ideas tho

8sunyuan: I was also iffy on my initial naming as its not exactly clear what is being checked. `staleStakesForbidden` sounds like a much better alternative but I'll ask for more input

8sunyuan: Also forgot as well its not timestamp based anymore, its based on blocks too

src/BLSSignatureChecker.sol
```diff
@@ -202,4 +220,13 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
...
     }
+
+    /**
+     * ServiceManager owner can either enforce or not that operator stakes are staler
+     * than the delegation.withdrawalDelayBlocks() window.
+     * @param value to toggle staleStakesForbidden
+     */
+    function setStaleStakesForbidden(bool value) external onlyServiceManagerOwner {
+        staleStakesForbidden = value;
+    }
```
wadealexc, line 231: Do we want an event for this?

src/RegistryCoordinator.sol
```diff
@@ -498,6 +541,29 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
+     * does nothing if operator is not registered for any quorums.
+     */
+    function _updateOperator(
+        address operator,
+        OperatorInfo memory operatorInfo,
+        bytes memory quorumsToUpdate
+    ) internal {
+        if (operatorInfo.status != OperatorStatus.REGISTERED) {
+            return;
+        }
```
wadealexc, line 555: I made an earlier comment that we should have this check for both methods and that it didn't cost extra because we're already SLOADing the operatorId

But I'm realizing that's actually not the same slot - the `operatorId` is 32 bytes, so the status is its own SLOAD.

If we want to save an extra SLOAD for each operator updated in `updateOperatorsForQuorum`, we could remove the `REGISTERED` check, put that solely in `updateOperators`, and just pass the `operatorId` into this helper method rather than the entire struct.

src/RegistryCoordinator.sol
```diff
@@ -257,31 +260,69 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
-                    quorumNumbers: BitmapUtils.bitmapToBytesArray(quorumsToRemove)
-                });    
+        for (uint256 i = 0; i < quorumNumbers.length; ++i) {
+            uint8 quorumNumber = uint8(quorumNumbers[i]);
+            address[] calldata currQuorumOperators = operatorsPerQuorum[i];
+            require(
+                currQuorumOperators.length == indexRegistry.totalOperatorsForQuorum(quorumNumber),
+                "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: number of updated operators does not match quorum total"
+            );
+            address prevOperatorAddress = address(0);
```
wadealexc, line 302: It doesn't seem this ever gets updated

src/RegistryCoordinator.sol
```diff
@@ -257,31 +260,69 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
+            // Update stakes for each operator in this quorum
+            for (uint256 j = 0; j < currQuorumOperators.length; ++j) {
+                address operator = currQuorumOperators[j];
+                OperatorInfo storage operatorInfo = _operatorInfo[operator];
```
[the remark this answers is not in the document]
8sunyuan: Updated https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/5dc29f626c488c4fe75ce6be2bad27dfd3514486

src/RegistryCoordinator.sol
```diff
@@ -257,31 +260,69 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
+     * @dev This method is used to update the stakes of all operators in a quorum at once, rather than individually. Performs
+     * sanitization checks on the input array lengths, quorumNumbers existing, and that quorumNumbers are ordered. Function must
+     * also not be paused by the PAUSED_UPDATE_OPERATOR flag.
+     */
+    function updateOperatorsForQuorum(
+        address[][] calldata operatorsPerQuorum,
+        bytes calldata quorumNumbers
+    ) external onlyWhenNotPaused(PAUSED_UPDATE_OPERATOR) {
+        uint192 quorumBitmap = uint192(BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers));
+        require(_quorumsAllExist(quorumBitmap), "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: some quorums do not exist");
```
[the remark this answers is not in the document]
8sunyuan: Fixed https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/5dc29f626c488c4fe75ce6be2bad27dfd3514486

src/RegistryCoordinator.sol
```diff
@@ -67,6 +67,9 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
     mapping(address => OperatorInfo) internal _operatorInfo;
     /// @notice whether the salt has been used for an operator churn approval
     mapping(bytes32 => bool) public isChurnApproverSaltUsed;
+    /// @notice mapping from quorum number to the latest block that all quorums were updated all at once
+    mapping(uint8 => uint256) public quorumUpdateBlockNumber;
```
gpsanant, line 71: blocknumber can be uint32 in case we want to store something more here in the future

src/BLSSignatureChecker.sol
```diff
@@ -202,4 +220,14 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
...
     }
+
+    /**
+     * ServiceManager owner can either enforce or not that operator stakes are staler
+     * than the delegation.withdrawalDelayBlocks() window.
+     * @param value to toggle staleStakesForbidden
+     */
+    function setStaleStakesForbidden(bool value) external onlyServiceManagerOwner {
+        staleStakesForbidden = value;
+        emit StaleStakesForbiddenUpdate(value);
```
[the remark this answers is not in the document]
wadealexc: why? it's public so it matches the getter

we're really inconsistent about this throughout both repos :sweat:

wadealexc (approved):
LGTM after fixes!

src/RegistryCoordinator.sol
8sunyuan: For weekly avs-sync, 200 operators in quorum0 and 60 operators in 9 remaining quorums, that's ~150,000 gas from the SLOADS. Maybe worth considering removing the check but I also prefer the explicitness

PR #65, opened 2023-11-16
update nonSignerForQuorumIndex outside of loop

gpsanant:
fixes a bug where the first nonSignerStakeIndex is used for all non signers


ChaoticWalrus (approved):
Good commit! We should add a regression test for this bug as well, but that can go in another PR.

PR #66, opened 2023-11-16
Anvil testing updateOperatorsForQuorum

8sunyuan:
Ran tests with anvil node running the following commands
```
anvil -f https://ethereum-goerli.publicnode.com --accounts 200
forge script script/test/M2_DeployTesting.s.sol:Deployer_M2 --rpc-url http://127.0.0.1:8545  --private-key $PRIVATE_KEY --broadcast -vvv
```

Assumptions:
- Quorum0(10 strategies) has 200 operators
- Quorums1-9(1 strategy each) has 50 operators each

Running updates for Quorum0 with 200 operators costs ~11.4M gas, updates for Quorums1-9 with 50 operators each costs ~6.4M gas. If using 30 operator cap instead of 50 then its ~4M. 

Only mocked the `BLSPubkeyRegistry` contract in tests as generating BLS pubkeys for all 200 operators would have been much more of a chore. This affects gas estimations in the event that the operator stake is below the minimum requirement for the quorum and we have to rederegister them from all other registries.
In this scenario, for Quorum0 and de-registering ALL 200 operators, it would cost approximately 2-3M more gas to perform the transaction. For updating all Quorums1-9 and all operators de-registering, it would cost approximately 3-5M more gas. 

**TL;DR** In the worst case scenario where all operators deregister/have below minimum quorum stake, we can still update operators for Quorum0 and Quorums1-9 in two separate tx's.


src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -502,6 +532,72 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                 */
+                if (!quorumsToRemove.isEmpty()) {
+                    _deregisterOperator({
+                        operator: operator,
+                        quorumNumbers: quorumNumbers[i:i+1]
+                    });    
+                }
+            }
+
+            quorumUpdateTimestamp[quorumNumber] = block.timestamp;
```
gpsanant, line 596: comment

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -502,6 +532,72 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+                address operator = currQuorumOperators[j];
+                OperatorInfo storage operatorInfo = _operatorInfo[operator];
+                bytes32 operatorId = operatorInfo.operatorId;
+                require(
+                    operatorInfo.status == OperatorStatus.REGISTERED,
+                    "BLSRegistryCoordinatorWithIndices.updateOperatorsForQuorum: Operator not registered"
+                );
+
+                {
+                    uint192 currentBitmap = _currentOperatorBitmap(operatorId);
```
gpsanant, line 563: any way to reduce duplication between this and the other update stakes method?

src/BLSSignatureChecker.sol
```diff
@@ -73,6 +73,10 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
         // loop through every quorumNumber and keep track of the apk
         BN254.G1Point memory apk = BN254.G1Point(0, 0);
         for (uint i = 0; i < quorumNumbers.length; i++) {
+            require(
+                registryCoordinator.quorumUpdateTimestamp(uint8(quorumNumbers[i])) + 1 weeks <= block.timestamp,
```
gpsanant, line 77: should we set a minimum withdrawalDelayBlocks in delegation manager and set this to that?

cc @ChaoticWalrus

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: Let me try to create another PR for this

src/BLSSignatureChecker.sol
ChaoticWalrus: what if we just read the current `DelegationManager.withdrawalDelayBlocks` value instead?

src/BLSRegistryCoordinatorWithIndices.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/68
Take a look here

src/BLSSignatureChecker.sol
gpsanant: oh yup! that'l work

8sunyuan: Added this in the feature PR https://github.com/Layr-Labs/eigenlayer-middleware/pull/62/commits/e4f39f61b962319578e8d8e4c0ef4582084833c7 as well as the interface change to read off of the DM https://github.com/Layr-Labs/eigenlayer-contracts/pull/343

PR #68, opened 2023-11-16
Refactor: RegistryCoordinator and Registry contracts

8sunyuan:
Made some refactoring changes as part of trying to reuse some code in `updateOperators` and `updateOperatorsForQuorum`

Moved the looping per `bytes memory quorumNumbers` into the registryCoordinator so that `deregisterOperator` for BLSPubkeyRegistry/IndexRegistry/StakeRegistry only takes a single quorum as an input. Makes it much cleaner when deregistering for a single quorum after doing a `stakeRegistry.updateOperatorStake()` within the  `updateOperators`/`updateOperatorsForQuorum` function since originally we were doing a bytes slicing anyways for a single quorum.
However does result in multiple external calls if registering/deregistering for multiple quorums.

**Note**: Only made changes for deregistering, if PR looks good then will make the same changes for registering

### DIFF
--- a/src/BLSPubkeyRegistry.sol
+++ b/src/BLSPubkeyRegistry.sol
@@ -30,17 +30,15 @@ contract BLSPubkeyRegistry is BLSPubkeyRegistryStorage {
     /**
      * @notice Registers the `operator`'s pubkey for the specified `quorumNumbers`.
      * @param operator The address of the operator to register.
-     * @param quorumNumbers The quorum numbers the operator is registering for, where each byte is an 8 bit integer quorumNumber.
+     * @param quorumNumbers The quorum number the operator is registering for
      * @return pubkeyHash of the operator's pubkey
      * @dev access restricted to the RegistryCoordinator
      * @dev Preconditions (these are assumed, not validated in this contract):
-     *         1) `quorumNumbers` has no duplicates
-     *         2) `quorumNumbers.length` != 0
-     *         3) `quorumNumbers` is ordered in ascending order
-     *         4) the operator is not already registered
+     *         1) the operator is not already registered
      */
     function registerOperator(
         address operator,
+        // uint8 quorumNumber
         bytes memory quorumNumbers
     ) public virtual onlyRegistryCoordinator returns (bytes32) {
         // Get the operator's pubkey from the compendium. Reverts if they have not registered a key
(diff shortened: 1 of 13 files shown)


gpsanant:
What's the benefit of this change?

8sunyuan:
> What's the benefit of this change?

Removing some redundancy. In a lot of places where we call `_deregisterOperator` its only being called for quorumNumbers[i:i+1] so its not really needed then to have additional for looping in the registry contracts. 
`registerOperatorWithChurn` for example is already doing checks for each quorum

src/BLSPubkeyRegistry.sol
```diff
@@ -126,6 +121,30 @@ contract BLSPubkeyRegistry is BLSPubkeyRegistryStorage {
         }
     }
 
+    function _processSingleQuorumApkUpdate(uint8 quorumNumber, BN254.G1Point memory point) internal {
+        uint256 historyLength = apkHistory[quorumNumber].length;
+        require(historyLength != 0, "BLSPubkeyRegistry._processQuorumApkUpdate: quorum does not exist");
```
ChaoticWalrus, line 126: ngl this feels like unnecessary code duplication. just using a length-1 array ain't that bad IMO / not worth the risk of missing something in duplication (especially if we ever modify the code)

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -264,31 +274,17 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
-
             address operator = operators[i];
+            // Get operator info and bitmap of quorums operator is registered for
             OperatorInfo storage operatorInfo = _operatorInfo[operator];
             bytes32 operatorId = operatorInfo.operatorId;
-
-            // Only update operators currently registered for at least one quorum
-            if (operatorInfo.status != OperatorStatus.REGISTERED) {
-                continue;
-            }
```
ChaoticWalrus, line 275: seems that by removing this we would now revert if any of the operators in the array is not registered, but perhaps we should confirm on that.
I liked this code since it deals ~fairly elegantly with the race condition where an operator de-registers while someone is trying to perform an update -- it would definitely not be ideal to have a large update (say 5 million gas) revert very late.

src/libraries/BitmapUtils.sol
```diff
@@ -300,6 +300,13 @@ library BitmapUtils {
         return (((bitmap >> numberToCheckForInclusion) & 1) == 1);
     }
 
+    /**
+     * @notice clear the `bitToClear` in `bitmap`
+     */
+    function clearBit(uint256 bitmap, uint8 bitToClear) internal pure returns (uint256) {
+        return bitmap & ~(1 << bitToClear);
+    }
```
ChaoticWalrus, line 308: I like this little additional helper 😄

ChaoticWalrus (commented):
So overall... IDK about this PR.
It seems -- fine? But probably unnecessary?
I feel like the major outcome is just a lot more looped external calls in the case where an operator is deregistered from multiple quorums at once.

I like some of the ideas here but think there's a simpler / less "invasive" way to make the `updateOperatorsForQuorum` function work in a less duplicative way, with ~the existing functions.

Like maybe you could:

- take the bulk of the `updateOperators` function and put it in a new, internal, `_updateOperators` or `_updateOperator` function (which you'd loop through, in the second case)
- perform a looped check for operator ordering and existence / registration for the quorum in question, all at the start of `updateOperatorsForQuorum`
- call the new `_updateOperator`(s) function after performing all these checks

8sunyuan:
> So overall... IDK about this PR. It seems -- fine? But probably unnecessary? I feel like the major outcome is just a lot more looped external calls in the case where an operator is deregistered from multiple quorums at once.
> 
> I like some of the ideas here but think there's a simpler / less "invasive" way to make the `updateOperatorsForQuorum` function work in a less duplicative way, with ~the existing functions.
> 
> Like maybe you could:
> 
> * take the bulk of the `updateOperators` function and put it in a new, internal, `_updateOperators` or `_updateOperator` function (which you'd loop through, in the second case)
> * perform a looped check for operator ordering and existence / registration for the quorum in question, all at the start of `updateOperatorsForQuorum`
> * call the new `_updateOperator`(s) function after performing all these checks
> 
> Happy to chat about ideas and try to figure out what makes the most sense here!

So without going through the changes in this draft, I moved some of the shared logic into `_updateOperator()` which we loop over for all the operators. I'm not sure how to implement your second point of "perform a looped check for operator ordering and existence" since that would require looping through all the operators first, and then having to loop through the same array of operators again for the `_updateOperator()` function. Seems it would be too gas intensive compared to one pass through currently. Here is the most recent implementation I wrote below, maybe this is kind of a middleground of the ideas https://github.com/Layr-Labs/eigenlayer-middleware/blob/e4f39f61b962319578e8d8e4c0ef4582084833c7/src/BLSRegistryCoordinatorWithIndices.sol#L288

PR #73, opened 2023-11-27
Operator<>AVS Mapping Support

ypatil12:
This draft PR adds functionality to support off-chain indexing for avs and operator info. Specifically, the funcationality enables support for the 5 use cases defined [here](https://docs.google.com/document/d/181iDl3inNaYi69ci4zpSFHkIzTEdUG3BPFAGmgZO-_E/edit). 

Summary of Changes:

- Functionality in RegistryCoordinator to call DelegationManager when an operator registers or is deregistered
- `UpdateAVSMetadataURI` function in RegistryCoordinator
- To support indexing of strategies supported by AVSs and restated strategies of operators, we define two functions: `getOperatorRestakedStrategies` and `getRestakeableStrategies`. These functions utilize quorum bitmaps of the AVS (new functionality) and operator respectively.


wadealexc:
Can you retarget the PR against `m2-mainnet`?

src/VoteWeigherBaseStorage.sol
```diff
@@ -39,6 +39,9 @@ abstract contract VoteWeigherBaseStorage is Initializable, IVoteWeigher {
     /// @notice The number of quorums that the VoteWeigher is considering.
     uint16 public quorumCount;
 
+    /// @notice Bitmap of quorums that are being used by the middleware.
+    uint192 public quorumBitmap;
```
gpsanant, line 43: would prefer to just use quorum count here?

src/StakeRegistry.sol
```diff
@@ -531,4 +531,67 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+     */
+    function getRestakeableStrategies() external view returns (IStrategy[] memory){
+        if(quorumBitmap == 0) {
+            return new IStrategy[](0);
+        }
+
+        bytes memory quorumNumbers = BitmapUtils.bitmapToBytesArray(quorumBitmap);
+        IStrategy[] memory strategies = new IStrategy[](_getNumStrategiesInBitmap(quorumNumbers));
+        uint256 index = 0;
+        for(uint256 i = 0; i < quorumNumbers.length; i++){
```
gpsanant, line 547: use quourmCount across the board

src/StakeRegistry.sol
```diff
@@ -531,4 +531,67 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+            StrategyAndWeightingMultiplier[] memory strategiesAndMultipliers = strategiesConsideredAndMultipliers[uint8(quorumNumbers[i])];
+            for(uint256 j = 0; j < strategiesAndMultipliers.length; j++){
+                strategies[index] = strategiesAndMultipliers[j].strategy;
```
gpsanant, line 550: this  might end up having a lot of duplicates. lets try to avoid that

src/StakeRegistry.sol
```diff
@@ -531,4 +531,67 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+                index++;
+            }
+        }
+
+        return strategies;
+    }
+
+    /**
+     * @notice Return a list of all strategies and corresponding share amounts for which the operator has restaked
+     * @dev There may strategies for which the operator has restaked on the quorum but has no shares. In this case,
```
gpsanant, line 560: @return

src/StakeRegistry.sol
```diff
@@ -531,4 +531,67 @@ contract StakeRegistry is VoteWeigherBase, StakeRegistryStorage {
...
+        for(uint256 i = 0; i < restakedQuorums.length; i++) {
+            StrategyAndWeightingMultiplier[] memory strategiesAndMultipliers = strategiesConsideredAndMultipliers[uint8(restakedQuorums[i])];
+            for (uint256 j = 0; j < strategiesAndMultipliers.length; j++) {
+                strategies[index] = strategiesAndMultipliers[j].strategy;
+                shares[index] = getCurrentOperatorStakeForQuorum(operatorId, uint8(restakedQuorums[i]));
+                index++;
+            }
+        }
+
+        return (strategies, shares);
```
gpsanant, line 587: i don't think it makes sense to return shares? let's just assume all stake of the strategy is staked until slashing is live

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -412,6 +432,8 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
                 status: OperatorStatus.REGISTERED
             });
 
+            delegationManager.registerOperatorWithAVS(operator, signatureWithSaltAndExpiry);
```
wadealexc, line 435: I wish we could skip this signature if the caller was the operator. Seems like unnecessary overhead, but I guess it makes sense in the context of untrusted AVSs...

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -381,8 +400,9 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
         address operator, 
         bytes32 operatorId,
         bytes calldata quorumNumbers,
-        string memory socket
-    ) internal virtual returns (RegisterResults memory) {
+        string memory socket,
+        SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry
```
wadealexc, line 404: Would prefer we rename this to `operatorSignature`. Less wordy and more clear intention.

src/StakeRegistry.sol
```diff
@@ -201,6 +202,8 @@ contract StakeRegistry is StakeRegistryStorage {
             nextUpdateBlockNumber: 0,
             stake: 0
         }));
+
+        quorumBitmap = uint192(quorumBitmap.plus(uint192(1) << quorumNumber));
```
wadealexc, line 206: Would prefer we add a helper for this, because it's easy to fuck up. I'm using this in another branch: 

https://github.com/Layr-Labs/eigenlayer-middleware/blob/a6baa25ef060b53a46dc7c77b769d8711fbe94dc/src/libraries/BitmapUtils.sol#L303-L311

src/StakeRegistry.sol
```diff
@@ -245,6 +248,11 @@ contract StakeRegistry is StakeRegistryStorage {
             _strategyParams[indicesToRemove[i]] = _strategyParams[_strategyParams.length - 1];
             _strategyParams.pop();
         }
+
+        // If no strategies exist in quorum, turn bit off 
+        if (_strategyParams.length == 0) {
+            quorumBitmap = uint192(quorumBitmap.minus(uint192(1) << quorumNumber));
```
wadealexc, line 254: Would prefer a helper for this. Could use something similar to the `setBit` method I linked, but with a bool on/off switch?

wadealexc: Or just `unsetBit` :shrug:

wadealexc (changes requested):
Can we remove the StakeRegistry changes in favor of exposing similar methods in the registry coordinator?

Specifically, I like how currently only one contract keeps track of the current set of initialized quorums, and I want to keep that logic there.

I think if you expose the for loops you're doing in both `view` `StakeRegistry` methods, the registry coordinator is a better place to keep this logic.

src/BLSRegistryCoordinatorWithIndices.sol
ChaoticWalrus, line 404: probably _also_ want to name this `operatorSignature` or similar, to make its usage a little clearer?

src/StakeRegistry.sol
```diff
@@ -704,4 +712,71 @@ contract StakeRegistry is StakeRegistryStorage {
...
+        return (strategies, shares);
+    }
+
+    function _getNumStrategiesInBitmap(bytes memory quorums) internal view returns (uint256) {
+        uint256 strategyCount;
+        for(uint256 i = 0; i < quorums.length; i++) {
+            strategyCount += strategyParams[uint8(quorums[i])].length;
+        }
+        return strategyCount;
+    }
```
ChaoticWalrus, line 781: quick thought:
if the same strategy is shared between any quorums, this will double-count it.
doesn't seem like a config anyone _should_ use, but this just occurred to me

I also think the use of `Bitmap` in this function name is a misnomer, since it's taking a bytes object as input

ChaoticWalrus (commented):
PR seems very reasonable, but would like to get the other PR this relies on just a bit more finalized first.
The new 'view' functions being added to the StakeRegistry seem ~inefficient, but AFAICT these are designed to only ever be called off-chain, and thus seem OK as-is.

src/StakeRegistry.sol
ypatil12: Ya, occurred to me as well. We have double counting where strategies are added twice and "null" counting where strategies are in a quorum and the operator hasn't restated for every strategy in said quorum. The indexing isn't perfect and the backend will end up de-duping and discarding in both cases

ypatil12: Ya, naming is stale, will update to bytes

ChaoticWalrus: OK, seems reasonable, probably want to add a dev tag or some other explanatory note though!
even some like just `@dev does not de-duplicate the list`

ChaoticWalrus:
> IIUC, we don't need this PR anymore in middle repo right? @ypatil12

We'd still need the parts that integrate with the new changes to the 'core' contracts, but we can delete the added functions in the StakeRegistry + associated interface from this PR.

bowenli86:
> > IIUC, we don't need this PR anymore in middle repo right? @ypatil12
> 
> We'd still need the parts that integrate with the new changes to the 'core' contracts, but we can delete the added functions in the StakeRegistry + associated interface from this PR.

ok, will work on a new pr for it

src/StakeRegistry.sol
```diff
@@ -19,7 +19,8 @@ import "src/StakeRegistryStorage.sol";
  * @author Layr Labs, Inc.
  */
 contract StakeRegistry is StakeRegistryStorage {
-    
+    using BitmapUtils for *;
```
gpsanant, line 22: hm. why not just uint256

wadealexc: doesn't make a difference to the bytecode :shrug:

ypatil12:
Closing this PR as functionality is out of scope.

src/interfaces/IRegistryCoordinator.sol
```diff
@@ -83,4 +85,19 @@ interface IRegistryCoordinator {
...
+     * @param signatureWithSaltAndExpiry is the signature of the operator used by the avs to register the operator in the delegation manager
+     */
+    function registerOperatorWithCoordinator(bytes memory quorumNumbers, bytes calldata registrationData, SignatureWithSaltAndExpiry memory signatureWithSaltAndExpiry) external;
+
+    /**
+     * @notice Deregisters the msg.sender as an operator from the middleware
+     * @param quorumNumbers are the bytes representing the quorum numbers that the operator is registered for
+     * @param deregistrationData is the the data that is decoded to get the operator's deregistration information
+     */
+    function deregisterOperatorWithCoordinator(bytes calldata quorumNumbers, bytes calldata deregistrationData) external;
```
bowenli86, line 102: QQ: why these 2 api are not implemented?

bowenli86:
see https://eigenlabs.slack.com/archives/C05PX94L3H9/p1702240195992199?thread_ts=1702146213.952949&cid=C05PX94L3H9

------

i really wanna do the middleware pr, but looking at my calendar - i'll be OOO on 14th, and hv a pile of backend work need to finish before then, realistically i won't be able to finish middleware pr then

will be great if @ypatil12 can continue the middleware pr

src/interfaces/IRegistryCoordinator.sol
wadealexc: they're implemented, just renamed `registerOperator` and `deregisterOperator`

PR #75, opened 2023-11-29
Update README.md to avoid confusing phrasing

bowenli86:
I initially didn't understand "Registry Coordinator/Registry Architecture" in the doc as this phrasing appears to be symmetric of "Registry Coordinator" + "Registry Architecture"

after reading the doc, i realized it's actually "Architecture of Registry Coordinator and Registry"

so making a change to avoid confusing others


ChaoticWalrus (approved):
idk if this is the ideal wording, but I approve of this change. Seems clearer at least.

PR #76, opened 2023-11-29
Optimize BLSSignatureChecker

0x0aa0:


src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -404,6 +407,16 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
             newBitmap: newBitmap
         });
 
+        //Update the operator contribution to aggregate public key across all quorums
+        BN254.G1Point memory operatorPubkey = blsPubkeyRegistry.getOperatorPubkey(operator);
```
gpsanant, line 411: internal function for this?

src/BLSRegistryCoordinatorWithIndices.sol
```diff
@@ -404,6 +407,16 @@ contract BLSRegistryCoordinatorWithIndices is EIP712, Initializable, IBLSRegistr
...
+        operatorNegativeAPKs[operatorId] = operatorNegativeAPKs[operatorId].plus(
```
gpsanant, line 412: we should get rid of the other "time series" pubkey and replace it with this

src/BLSSignatureChecker.sol
```diff
@@ -69,96 +69,81 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
...
-                            )
+                uint256 nonSignerQuorumBitmap = 
+                    registryCoordinator.getQuorumBitmapAtBlockNumberByIndex(
+                        nonSignerPubkeyHashes[i], 
+                        referenceBlockNumber, 
+                        nonSignerStakesAndSignature.nonSignerQuorumBitmapIndices[i]
                     );
+                
+                {
+                    BN254.G1Point memory operatorNegativeAPK = registryCoordinator.getOperatorNegativeAPK(nonSignerPubkeyHashes[i]);
```
gpsanant, line 113: this needs to be verified to be at the reference block number

src/BLSSignatureChecker.sol
```diff
@@ -168,10 +146,8 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
             require(pairingSuccessful, "BLSSignatureChecker.checkSignatures: pairing precompile call failed");
             require(signatureIsValid, "BLSSignatureChecker.checkSignatures: signature is invalid");
         }
-        // set signatoryRecordHash variable used for fraudproofs
```
gpsanant, line 171: why deleting comments

src/BLSSignatureChecker.sol
```diff
@@ -69,96 +69,74 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
             bytes32
         )
     {   
-        // verify the provided apk was the apk at referenceBlockNumber
-        // loop through every quorumNumber and keep track of the apk
+        QuorumStakeTotals memory quorumStakeTotals;
```
gpsanant, line 72: do we have estimates of how much gas this saves

PR #79, opened 2023-11-30
chore: merge master into m2-mainnet

wadealexc:
a rebase would be better, but i don't have the time, so (ancestors forgive me) i'm merging instead


wadealexc:
> LGTM, although did we get around to writing the regression test for that index bug in the signature checker?

No we have not, but we havent gotten around to a lot of testing items for this repo :sweat:

PR #80, opened 2023-12-01
Build: Set lib/eigenlayer-contracts submodule to m2-mainnet branch

8sunyuan:
Updating commit from `bccae53` to `b91a743`
Latest m2-mainnet commits can be viewed here https://github.com/Layr-Labs/eigenlayer-contracts/commits/m2-mainnet


ChaoticWalrus (approved):
wow, no other changes necessary?
that's 🔥 
good call with this

PR #85, opened 2023-12-04
feat: more general interface for querying operator weights

ChaoticWalrus:
We'd like to have more flexible, forward-looking interfaces. Constraining the sizes of input/return variables _in the interface itself_ doesn't seem to be serving any practical purpose here, so this commit relaxes the requirement.

Basically no cost / slightly more flexibility :+1:

### DIFF
--- a/src/StakeRegistry.sol
+++ b/src/StakeRegistry.sol
@@ -495,14 +495,14 @@ contract StakeRegistry is StakeRegistryStorage {
     *******************************************************************************/
 
     /**
-     * @notice This function computes the total weight of the @param operator in the quorum @param quorumNumber.
+     * @notice This function computes the total weight of the @param operator in the quorum @param quorumId.
      * @dev reverts if the quorum does not exist
      */
-    function weightOfOperatorForQuorum(
-        uint8 quorumNumber, 
-        address operator
-    ) public virtual view quorumExists(quorumNumber) returns (uint96) {
-        (uint96 stake, ) = _weightOfOperatorForQuorum(quorumNumber, operator);
+    function weightOfOperator(
+        address operator,
+        bytes32 quorumId
+    ) public virtual view quorumExists(uint8(uint256(quorumId))) returns (uint256) {
+        (uint96 stake, ) = _weightOfOperatorForQuorum(uint8(uint256(quorumId)), operator);
         return stake;
     }
 
--- a/src/interfaces/IStakeRegistry.sol
+++ b/src/interfaces/IStakeRegistry.sol
@@ -140,11 +140,8 @@ interface IStakeRegistry is IRegistry {
         uint256 index
     ) external view returns (StrategyParams memory);
 
-    /**
-     * @notice This function computes the total weight of the @param operator in the quorum @param quorumNumber.
-     * @dev reverts in the case that `quorumNumber` is greater than or equal to `quorumCount`
-     */
-    function weightOfOperatorForQuorum(uint8 quorumNumber, address operator) external view returns (uint96);
+    // @notice This function computes the weight of the @param operator in the quorum @param quorumId.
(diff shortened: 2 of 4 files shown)


wadealexc:
The changes themselves look fine - but I'm curious why we're not changing more than just this one function. Do we want to use `bytes32` consistently, or just in this method? If the latter, why?

ChaoticWalrus:
> The changes themselves look fine - but I'm curious why we're not changing more than just this one function. Do we want to use `bytes32` consistently, or just in this method? If the latter, why?

good question!
This seemed like the function most likely to get "generalized" / fit into a wider "AVS interface" -- it feels the most fundamental + like what external actors would be the most interested in, and in particular relaxing the input/return _sizes_ was my primary concern. @gpsanant wanted bytes32 instead of uint256 (because he believed this would be easier on the off-chain side), and I didn't have a strong opinion so trusted that thought.

That being said, I do think we should probably modify more here, but am not sure on the extent.
I think making the events more general (so changing uint8 quorumNumbers to bytes32 quorumIds and uint96 minimumStake=>uint256) could be good as well.

I'm somewhat more hesitant to change all of the interface, as we use uint8's for storage packing in some places, the contracts use each others' interfaces quite a bit (so it'd end up being a systemic change to the whole set of Registry contracts, I think) + I find it to be semantically clearer to use uint8s in contexts where we are both (a) using these explicitly as numbers, rather than a more abstract "id" and (b) restricting the size to fit within uint8 (technically even smaller).

We could always provide _both_ functions? I don't think I'm super opposed, just feels a bit "excessive" perhaps?

@wadealexc I requested your review in particular because I think you tend to offer good perspective on these types of decisions.

wadealexc:
I definitely don't want to provide both; I think we should pick a side and go with it.

But I feel like changing just this one function is... unecessary? We want to be more future compatible, I think, is the motivation?

But future compatibility is most affected by the contract's storage, not its interface. So I don't understand why we'd make this change now when fully supporting bytes32 quorums would require a whole ass upgrade down the line.

wadealexc:
It's just really weird to do only this one function. I'm not opposed to a more generalizable quorum id - as long as we move everything to that and not just this bit.

ChaoticWalrus:
> It's just really weird to do only this one function. I'm not opposed to a more generalizable quorum id - as long as we move everything to that and not just this bit.

fair enough.

> We want to be more future compatible, I think, is the motivation?

yeah, I think that's a good way of summing up the motivation. since people will be building on these defaults or modelling their own work on them, it'd help to be more general / flexible here. But I agree only doing this for one function doesn't really accomplish that. I wanted there to be a middle-ground here but I think if there is one, it presumably involves changing a bit more.

> But future compatibility is most affected by the contract's storage, not its interface. So I don't understand why we'd make this change now when fully supporting bytes32 quorums would require a whole ass upgrade down the line.

So my thought process at least was that if we set a more general interface, we offer greater future-compatibility / avoid handicapping other developers trying to match an interface, even if our usage is "limited".
Yes supporting additional quorums (more than 192) would require an upgrade, and I don't want to take on the ambitious work of sorting that out at the moment. Our storage model is tied to this number through various optimizations, and for the use-cases we've identified, it seems sufficient. But I'd like for us to avoid unnecessarily constraining others, or necessitating an interface upgrade for ourselves _and them_ to match some new spec.

But this may be getting too ahead of things, as I think we'll all manage if we ship what we have already.
I anticipate requiring _some kind of upgrade_ regardless, and we will have a better idea of what that should look like + what the requirements are as we draw closer; trying to anticipate things right now is very difficult and we necessarily can't anticipate everything.
I don't want to rush changes through. Will give this a little more thought but I remain OK with dropping it and revisiting later. There will undoubtedly be costs to any path we take here 🤷

PR #95, opened 2023-12-06
chore: rename several contracts and fix interfaces and tests

wadealexc:
- BLSRegistryCoordinatorWithIndices is now RegistryCoordinator
- BLSOperatorStateRetriever is now OperatorStateRetriever
- BLSPubkeyRegistry is now BLSApkRegistry


src/interfaces/IRegistryCoordinator.sol
```diff
@@ -1,18 +1,29 @@
...
 
     /// Emits when an operator is deregistered
     event OperatorDeregistered(address indexed operator, bytes32 indexed operatorId);
-    
+
+    event OperatorSetParamsUpdated(uint8 indexed quorumNumber, OperatorSetParam operatorSetParams);
+
+    event ChurnApproverUpdated(address prevChurnApprover, address newChurnApprover);
+
+    event EjectorUpdated(address prevEjector, address newEjector);
```
ChaoticWalrus, line 25: ahhh, so the key change being made here is merging the `IBLSRegistryCoordinatorWithIndices` interface _into the existing_ `IRegistryCoordinator` interface.
Not sure if I understood that to be an intention of the PR from our discussion.
I'll review the state of the repo to better understand if I agree with this particular move.

ChaoticWalrus: On review, I'm OK with this. Seems the things in the interface aren't at all BLS-specific, other than the `blsApkRegistry`, and that's super reasonable to change in the future.

wadealexc: Definitely wasn't the intent, but seemed the easiest way to do things.

It's not a flawless BLS-neutral interface, but as you said, we can fix that later.

PR #97, opened 2023-12-07
Feat: adjust compendium for per-avs deployment

ChaoticWalrus:
Using a shared BLSPubkeyCompendium impacts separation of concerns, so we would like to transition to operators registering a public key with each AVS independently.
This means we can absorb the BLSPubkeyCompendium into the BLSApkRegistry (possibly re-re-named to BLSPubkeyRegistry?), which this PR looks to accomplish.

~~PR initially marked as draft; will request review when I think it's ready.~~ See later updates / I edited this message and crossed out some items.

Open items + open questions:

~~- fix tests (marked a few things with TODOs)~~
~~- decide on combined 'register operator + register pubkey' flow / function~~ (completed / merged into this PR in https://github.com/Layr-Labs/eigenlayer-middleware/pull/100)
~~- renaming functions / cleaning up interfaces~~ (completed in https://github.com/Layr-Labs/eigenlayer-middleware/pull/102)
~~- settle the message hash for operators -- see discussion here https://github.com/Layr-Labs/eigenlayer-middleware/pull/97#discussion_r1423155467 and here https://github.com/Layr-Labs/eigenlayer-middleware/pull/100#discussion_r1424472058~~ (completed in https://github.com/Layr-Labs/eigenlayer-middleware/pull/102)
~~- fix contract code size limits issue arising after resolving merge conflicts~~ (completed in https://github.com/Layr-Labs/eigenlayer-middleware/pull/105)


ChaoticWalrus:
OK, tests are now passing once again.
Updating with my current thoughts -- I am now planning to split renaming + the combined flow into _separate PRs based off of this branch_, so we can break things into more discrete parts and make some of those decisions non-blocking for merging others.
If you are reviewing this and disagree with this approach, let me know!

src/BLSApkRegistry.sol
```diff
@@ -148,6 +200,34 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
...
+    /**
+     * @notice Returns the message hash that an operator must sign to register their BLS public key.
+     * @param operator is the address of the operator registering their BLS public key
+     */
+    function getMessageHash(address operator) public view returns (BN254.G1Point memory) {
+        return BN254.hashToG1(keccak256(abi.encodePacked(
+            operator, 
+            address(this),
+            block.chainid, 
+            "EigenLayer_BN254_Pubkey_Registration"
```
8sunyuan, line 228: Should probably change the message hash "EigenLayer_BN254_Pubkey_Registration" to something that's more AVS specific, or just remove entirely since there won't be a singleton compendium contract anymore

ChaoticWalrus: hmm, perhaps we can just delete the `EigenLayer_` part?
it's overkill to have this at all, but it's another input to help protect from any kind of replay issues, so...I feel like it's not a bad idea to have _some_ string here at least.

ChaoticWalrus: realized I should probably just convert this to use EIP-712 instead; will add that to my TODOs

wadealexc (commented):
This looks great -- pulling out from Slack: 

My suggestion for adding pubkey registration to the workflow is to:
1. add a new method to the reg coord: `registerOperatorAndPubkey` 
2. change `ApkRegistry.registerBLSPublicKey` to be called by reg coord only
3. change `ApkRegistry.registerBLSPublicKey` to accept a param for the operator address rather than using msg.sender.

`registerOperatorAndPubkey` works just like `registerOperator`, except it includes the pubkey params needed for the ApkRegistry. It fails if the operator already has a registered pubkey.

wadealexc (approved):
Since I've approved your other chore/cleanup PR, i'm going to go ahead and approve this one as well - so you can merge em all at once (assuming you're happy)

I reviewed the additional changes to operatorId fetching from the last PR, and as mentioned the cleanup PR looks great!

Thanks for putting these together :smile:

src/BLSApkRegistry.sol
ChaoticWalrus: addressed as part of https://github.com/Layr-Labs/eigenlayer-middleware/pull/102 -- I think these changes are quite appropriate

ChaoticWalrus:
Summary of key changes ultimately contained in this PR:

- merge BLSPubkeyCompendium into BLSApkRegistry
- define a `PubkeyRegistrationParams` struct in the IBLSApkRegistry interface, and add it as an input to the `RegistryCoordinator.registerOperator` and `RegistryCoordinator.registerOperatorWithChurn` functions -- operators now make a single call to these functions instead of needing to make a separate, earlier call to register a pubkey -- this input is ignored if the caller (the operator) has already registered a pubkey
- better define the message hash that operators sign when registering a pubkey, and move the definition of this message hash from the compendium to the RegistryCoordinator

PR #98, opened 2023-12-11
chore: remove ServiceManagerBase and add RegistryCoordinator owner

wadealexc:
Also noticed that we did not have `_disableInitializers` in the reg coord constructor, so I added that.


src/interfaces/IBLSSignatureChecker.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity =0.8.12;
 
 import {IRegistryCoordinator} from "src/interfaces/IRegistryCoordinator.sol";
 import {IBLSApkRegistry} from "src/interfaces/IBLSApkRegistry.sol";
-import {IStakeRegistry, IDelegationManager, IServiceManager} from "src/interfaces/IStakeRegistry.sol";
+import {IStakeRegistry, IDelegationManager} from "src/interfaces/IStakeRegistry.sol";
```
ypatil12, line 6: Nit: The natspec for the event `StaleStakesForbiddenUpdate` notes that the value is set by the serviceManagerOwner

ypatil12 (approved):
LGTM, one minor nit

src/interfaces/IBLSSignatureChecker.sol
wadealexc: [da8f669](https://github.com/Layr-Labs/eigenlayer-middleware/pull/98/commits/da8f669aa139569306a427c1eac0e9410a5852ce)

PR #99, opened 2023-12-11
Feat: Add AVS/Operator Registration Support in RegistryCoordinator

ypatil12:
This PR adds AVS<>Operator registration support in the RegistryCoordinator. It calls the core contract functions added in [this core PR.](https://github.com/Layr-Labs/eigenlayer-contracts/pull/363)

Removed kickIndex in the `registerOperatorWithChurn` function as it was an unused variable.

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -9,6 +9,7 @@ import {EIP1271SignatureUtils} from "eigenlayer-contracts/src/contracts/librarie
 import {IPauserRegistry} from "eigenlayer-contracts/src/contracts/interfaces/IPauserRegistry.sol";
 import {Pausable} from "eigenlayer-contracts/src/contracts/permissions/Pausable.sol";
 import {ISlasher} from "eigenlayer-contracts/src/contracts/interfaces/ISlasher.sol";
+import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
 
 import {IRegistryCoordinator} from "src/interfaces/IRegistryCoordinator.sol";
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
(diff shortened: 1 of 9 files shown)


src/RegistryCoordinator.sol
```diff
@@ -209,10 +220,10 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
             operator: msg.sender,
             operatorId: operatorId,
             quorumNumbers: quorumNumbers,
-            socket: socket
+            socket: socket,
+            operatorSignature: operatorSignature
         });
 
-        uint256 kickIndex = 0;
```
wadealexc, line 215: Ah. This should stay in. `kickIndex` should be used to access `operatorKickParams` when calling `_validateChurn`. :expressionless:

wadealexc: And `_deregisterOperator` below that

src/RegistryCoordinator.sol
```diff
@@ -401,6 +411,15 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
     }
 
+    /**
+     * @notice Sets the metadata URI for the AVS
+     * @param _metadataURI is the metadata URI for the AVS
+     * @dev only callable by the service manager owner
+     */
+    function setMetadataURI(string memory _metadataURI) external onlyServiceManagerOwner {
+        delegationManager.updateAVSMetadataURI(_metadataURI);
+    }
```
wadealexc, line 421: Just need to rebase since the service manager was removed :)

wadealexc (changes requested):
Need to fix the bug I mentioned - but looks great otherwise!

(if you have time, could you add a regression test? I'm pretty sure we've had this exact issue before; not sure how it snuck back in :grimacing: )

src/RegistryCoordinator.sol
wadealexc: Nevermind - the lists are the same length! Good callout @ypatil12 :heart:

wadealexc (approved):
Looks great!

PR #100, opened 2023-12-12
Feat: combined operator registration flow

ChaoticWalrus:
Opening this is a "sub-PR" / add-on to https://github.com/Layr-Labs/eigenlayer-middleware/pull/97

Now that the pubkey compendium has been merged into other middleware contracts / is on a per-AVS basis, we can make the call to register a new pubkey permissioned + make a single function that combines what was previously 2 calls.

### DIFF
--- a/src/BLSApkRegistry.sol
+++ b/src/BLSApkRegistry.sol
@@ -94,22 +94,24 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
     }
 
     /**
-     * @notice Called by an operator to register themselves as the owner of a BLS public key and reveal their G1 and G2 public key.
+     * @notice Called by the RegistryCoordinator register an operator as the owner of a BLS public key.
+     * @param operator is the operator for whom the key is being registered
      * @param signedMessageHash is the registration message hash signed by the private key of the operator
      * @param pubkeyG1 is the corresponding G1 public key of the operator 
      * @param pubkeyG2 is the corresponding G2 public key of the operator
      */
     function registerBLSPublicKey(
+        address operator,
         BN254.G1Point memory signedMessageHash, 
         BN254.G1Point memory pubkeyG1, 
         BN254.G2Point memory pubkeyG2
-    ) external {
+    ) external onlyRegistryCoordinator {
         bytes32 pubkeyHash = BN254.hashG1Point(pubkeyG1);
         require(
             pubkeyHash != ZERO_PK_HASH, "BLSApkRegistry.registerBLSPublicKey: cannot register zero pubkey"
         );
         require(
-            operatorToPubkeyHash[msg.sender] == bytes32(0),
+            operatorToPubkeyHash[operator] == bytes32(0),
             "BLSApkRegistry.registerBLSPublicKey: operator already registered pubkey"
         );
         require(
@@ -118,7 +120,7 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
         );
 
         // H(m) 
-        BN254.G1Point memory messageHash = getMessageHash(msg.sender);
+        BN254.G1Point memory messageHash = getMessageHash(operator);
 
         // gamma = h(sigma, P, P', H(m))
         uint256 gamma = uint256(keccak256(abi.encodePacked(
@@ -140,11 +142,11 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
             pubkeyG2
         ), "BLSApkRegistry.registerBLSPublicKey: either the G1 signature is wrong, or G1 and G2 private key do not match");
 
-        operatorToPubkey[msg.sender] = pubkeyG1;
-        operatorToPubkeyHash[msg.sender] = pubkeyHash;
-        pubkeyHashToOperator[pubkeyHash] = msg.sender;
+        operatorToPubkey[operator] = pubkeyG1;
+        operatorToPubkeyHash[operator] = pubkeyHash;
+        pubkeyHashToOperator[pubkeyHash] = operator;
 
-        emit NewPubkeyRegistration(msg.sender, pubkeyG1, pubkeyG2);
+        emit NewPubkeyRegistration(operator, pubkeyG1, pubkeyG2);
     }
 
     /*******************************************************************************
--- a/src/BLSApkRegistry.sol
+++ b/src/BLSApkRegistry.sol
@@ -96,17 +96,13 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
     /**
      * @notice Called by the RegistryCoordinator register an operator as the owner of a BLS public key.
      * @param operator is the operator for whom the key is being registered
-     * @param signedMessageHash is the registration message hash signed by the private key of the operator
-     * @param pubkeyG1 is the corresponding G1 public key of the operator 
-     * @param pubkeyG2 is the corresponding G2 public key of the operator
+     * @param pubkeyRegistrationParams contains the G1 & G2 public keys of the operator, and a signature proving their ownership
      */
     function registerBLSPublicKey(
         address operator,
-        BN254.G1Point memory signedMessageHash, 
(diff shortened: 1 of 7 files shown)


src/BLSApkRegistry.sol
```diff
@@ -94,22 +94,20 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
...
+     * @param pubkeyRegistrationParams contains the G1 & G2 public keys of the operator, and a signature proving their ownership
      */
     function registerBLSPublicKey(
-        BN254.G1Point memory signedMessageHash, 
-        BN254.G1Point memory pubkeyG1, 
-        BN254.G2Point memory pubkeyG2
-    ) external {
-        bytes32 pubkeyHash = BN254.hashG1Point(pubkeyG1);
+        address operator,
+        PubkeyRegistrationParams calldata pubkeyRegistrationParams
```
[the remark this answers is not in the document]
wadealexc: clarifying: by "things" I mean the variable name, not the struct name

src/BLSApkRegistry.sol
```diff
@@ -118,33 +116,33 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
         );
 
         // H(m) 
-        BN254.G1Point memory messageHash = getMessageHash(msg.sender);
+        BN254.G1Point memory messageHash = getMessageHash(operator);
```
wadealexc, line 119: I wonder if the message hash should use the registry coordinator rather than the BLSApkRegistry, since the registry coordinator is "the AVS" (as far as the Dmgr is concerned, anyway)

wadealexc: i guess we're registering within the context of the AVS though so maybe it makes sense as-is

src/RegistryCoordinator.sol
```diff
@@ -145,11 +147,21 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
+        IBLSApkRegistry.PubkeyRegistrationParams calldata pubkeyRegistrationParams
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
+        /**
+         * IF the operator has never registered a pubkey before, THEN register their pubkey
+         * OTHERWISE, simply ignore the provided `pubkeyRegistrationParams`
+         */
+        if (blsApkRegistry.operatorToPubkeyHash(msg.sender) == 0) {
+            blsApkRegistry.registerBLSPublicKey(msg.sender, pubkeyRegistrationParams);
+        }
         bytes32 operatorId = blsApkRegistry.getOperatorId(msg.sender);
```
wadealexc, line 165: Since `getOperatorId` returns `operatorToPubkeyHash`, I think this is cleaner like this:

```
bytes32 operatorId = blsApkRegistry.getOperatorId(msg.sender);
if (operatorId == 0) {
  operatorId = blsApkRegistry.registerBLSPublicKey(msg.sender, params);
}
```

src/RegistryCoordinator.sol
```diff
@@ -180,18 +192,28 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
...
         OperatorKickParam[] calldata operatorKickParams,
         SignatureWithSaltAndExpiry memory churnApproverSignature
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
         require(operatorKickParams.length == quorumNumbers.length, "RegistryCoordinator.registerOperatorWithChurn: input length mismatch");
 
+        /**
+         * IF the operator has never registered a pubkey before, THEN register their pubkey
+         * OTHERWISE, simply ignore the `pubkeyRegistrationSignature`, `pubkeyG1`, and `pubkeyG2` inputs
+         */
+        if (blsApkRegistry.operatorToPubkeyHash(msg.sender) == 0) {
```
wadealexc, line 214: Same comment as above

src/RegistryCoordinator.sol
```diff
@@ -211,17 +233,16 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
 
         uint256 kickIndex = 0;
         for (uint256 i = 0; i < quorumNumbers.length; i++) {
-            uint8 quorumNumber = uint8(quorumNumbers[i]);
-            
-            OperatorSetParam memory operatorSetParams = _quorumParams[quorumNumber];
+            // reference: uint8 quorumNumber = uint8(quorumNumbers[i]);
+            OperatorSetParam memory operatorSetParams = _quorumParams[uint8(quorumNumbers[i])];
```
wadealexc, line 237: Ah, is this avoiding "stack too deep"?

You may be able to avoid this change -- Yash discovered that we're not actually using `kickIndex` and has removed it in this PR: https://github.com/Layr-Labs/eigenlayer-middleware/pull/99

wadealexc (approved):
Minor comments, nothing blocking

src/RegistryCoordinator.sol
ChaoticWalrus: hmm OK. yes this was the fix to my stack-too-deep issue

wadealexc: I checked on this when chatting with Yash.

Previously (check master) - we didn't require the input length arrays to be the same size. Now, we do - so the index can just be `i`

Personally I like this better - it means a handful of the kick params could be empty, but it's less complicated than keeping 2 loop indices.

ChaoticWalrus: ahhhh OK makes sense!
will delete this and revert the commented-out bit

src/RegistryCoordinator.sol
ChaoticWalrus: agreed, don't know why I thought this reverted if the operator wasn't registered or something. will make this change

src/BLSApkRegistry.sol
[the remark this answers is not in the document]
ChaoticWalrus: yeah fair enough. I think the variable is clear enough from context to shorten the name. making it happen.

src/BLSApkRegistry.sol
ChaoticWalrus: > i guess we're registering within the context of the AVS though so maybe it makes sense as-is

yeah I agree with this.
will make the message signature more EIP712-compatible, but I think using the BLSApkRegistry address here is OK

src/RegistryCoordinator.sol
ChaoticWalrus: deleted here https://github.com/Layr-Labs/eigenlayer-middleware/pull/100/commits/9948e01a7701e1f9f4ea37e83ba3e9d4343bd852

src/RegistryCoordinator.sol
ChaoticWalrus: changed here. good callout https://github.com/Layr-Labs/eigenlayer-middleware/pull/100/commits/db000925f55ba3bca2db005008fb55feab68e60b

src/RegistryCoordinator.sol
ChaoticWalrus: changed here. good callout https://github.com/Layr-Labs/eigenlayer-middleware/pull/100/commits/db000925f55ba3bca2db005008fb55feab68e60b

src/BLSApkRegistry.sol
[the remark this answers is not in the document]
ChaoticWalrus: renaming done here https://github.com/Layr-Labs/eigenlayer-middleware/pull/100/commits/da59eee8a66a1d83895dfc5994203ecb903823f9

ChaoticWalrus:
merging this PR
will note that this thread on the message hash for operators to sign is still open https://github.com/Layr-Labs/eigenlayer-middleware/pull/100#discussion_r1424472058
planning to address that as part of the "main" PR here (https://github.com/Layr-Labs/eigenlayer-middleware/pull/97)

PR #102, opened 2023-12-12
Chore: cleanup for compendium changes

ChaoticWalrus:
Aimed at addressing the last couple TODOs in https://github.com/Layr-Labs/eigenlayer-middleware/pull/97
I pulled these changes out into a separate PR because review has already been done on the existing changes in that PR / I think they are quite well-contained.
I initially thought I'd want to do more renaming, but I think the primary commit here -- https://github.com/Layr-Labs/eigenlayer-middleware/commit/630b27d394191b89d6c88e7a1b8dda07750a31db -- takes care of the worst offender, and I'm not strongly opinionated enough to fight for other naming changes right now.

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -35,6 +35,8 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
     /// @notice The EIP-712 typehash for the `DelegationApproval` struct used by the contract
     bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
         keccak256("OperatorChurnApproval(bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams)OperatorKickParam(address operator,bytes32[] operatorIdsToSwap)");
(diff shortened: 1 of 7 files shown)


src/RegistryCoordinator.sol
```diff
@@ -213,7 +215,7 @@ contract RegistryCoordinator is EIP712, Initializable, IRegistryCoordinator, ISo
          */
         bytes32 operatorId = blsApkRegistry.getOperatorId(msg.sender);
         if (operatorId == 0) {
-            operatorId = blsApkRegistry.registerBLSPublicKey(msg.sender, params);
+            operatorId = blsApkRegistry.registerBLSPublicKey(msg.sender, params, pubkeyRegistrationMessageHash(msg.sender));
         }
```
wadealexc, line 219: Could consider tossing these lines into a helper, since we call them twice.

`bytes32 operatorId = _getOrCreateOperatorId(msg.sender, params);`

I don't feel strongly about this though.

wadealexc (approved):
Looks great!

PR #104, opened 2023-12-13
fix: address multiple issues in BLSSignatureChecker, optimize, and clean

wadealexc:
Fixes:
- swap sign on withdrawal delay check
- adjust withdrawal delay check so that it checks against `referenceBlockNumber` rather than `block.number`
- add fix for `BN254.scalar_mul_tiny` from #76 

Other changes:
- comments, clarity, and renaming
- adds input validation for all parameters
- cache state to avoid lookups in loop
- remove an uneeded loop from checkSignatures
- only negate pubkeys a single time, rather than each time a nonsigner pk is added

### DIFF
--- a/src/BLSApkRegistry.sol
+++ b/src/BLSApkRegistry.sol
@@ -254,7 +254,6 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
          * - blockNumber should be >= the update block number
          * - the next update block number should be either 0 or strictly greater than blockNumber
          */
-        // TODO - should this fail if quorumApkUpdate.apkHash == 0? This will be the case for the first entry in each quorum
         require(
             blockNumber >= quorumApkUpdate.updateBlockNumber,
             "BLSApkRegistry._validateApkHashAtBlockNumber: index too recent"
--- a/src/BLSApkRegistry.sol
+++ b/src/BLSApkRegistry.sol
@@ -177,22 +177,6 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
         }
     }
 
-    // TODO - should this fail if apkUpdate.apkHash == 0? This will be the case for the first entry in each quorum
-    function _validateApkHashAtBlockNumber(ApkUpdate memory apkUpdate, uint32 blockNumber) internal pure {
-        require(
-            blockNumber >= apkUpdate.updateBlockNumber,
-            "BLSApkRegistry._validateApkHashAtBlockNumber: index too recent"
-        );
-        /**
-         * if there is a next update, check that the blockNumber is before the next update or if
-         * there is no next update, then apkUpdate.nextUpdateBlockNumber is 0.
-         */
-        require(
-            apkUpdate.nextUpdateBlockNumber == 0 || blockNumber < apkUpdate.nextUpdateBlockNumber,
-            "BLSApkRegistry._validateApkHashAtBlockNumber: not latest apk update"
-        );
-    }
-
     /*******************************************************************************
                             VIEW FUNCTIONS
     *******************************************************************************/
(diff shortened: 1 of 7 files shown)


src/BLSApkRegistry.sol
```diff
@@ -201,7 +185,22 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
...
     ) external view returns (bytes24) {
         ApkUpdate memory quorumApkUpdate = apkHistory[quorumNumber][index];
-        _validateApkHashAtBlockNumber(quorumApkUpdate, blockNumber);
+
+        /**
+         * Validate that the update is valid for the given blockNumber:
+         * - blockNumber should be >= the update block number
+         * - the next update block number should be either 0 or strictly greater than blockNumber
+         */
+        // TODO - should this fail if quorumApkUpdate.apkHash == 0? This will be the case for the first entry in each quorum
```
gpsanant, line 194: it's fine? other checks in the sig checker will fail if no operators have registered for the quorum

can add if it feels safer tho

src/RegistryCoordinator.sol
```diff
@@ -785,16 +785,22 @@ contract RegistryCoordinator is
         uint256 index
     ) external view returns (uint192) {
         QuorumBitmapUpdate memory quorumBitmapUpdate = _operatorBitmapHistory[operatorId][index];
+        
+        /**
+         * Validate that the update is valid for the given blockNumber:
+         * - blockNumber should be >= the update block number
+         * - the next update block number should be either 0 or strictly greater than blockNumber
+         */
+        // TODO - should this fail if quorumBitmapUpdate.quorumBitmap == 0? This will be the case for the first entry in each quorum
```
gpsanant, line 794: does this todo need to be here?

wadealexc: yeah i added one for each place we query in this manner

wanted the reviewers to chat about it and make sure we're happy with the status quo before i go removing each one

src/BLSSignatureChecker.sol
```diff
@@ -72,135 +77,178 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
...
+            uint256 signingQuorumBitmap = BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers, registryCoordinator.quorumCount());
+
+            for (uint256 j = 0; j < params.nonSignerPubkeys.length; j++) {
+                // The nonsigner's pubkey hash doubles as their operatorId
+                // The check below validates that these operatorIds are sorted (and therefore
+                // free of duplicates)
+                nonSigners.pubkeyHashes[j] = params.nonSignerPubkeys[j].hashG1Point();
+                if (j != 0) {
+                    require(
+                        uint256(nonSigners.pubkeyHashes[j]) > uint256(nonSigners.pubkeyHashes[j - 1]),
```
gpsanant, line 143: @0x0aa0 pls make sure that the offchain code can deal with this

in fact, we should document all interfaces breaking changes since m2 testnet so we know how everything must change in the clients

wadealexc: FWIW i did not add this - it was already in this method

0x0aa0: the DASM on testnet also has this sorting check so I do not think it is breaking

src/libraries/BN254.sol
```diff
@@ -268,10 +268,14 @@ library BN254 {
...
+    /// @return hashedG1 the keccak256 hash of the G1 Point
     /// @dev used for BLS signatures
-    function hashG1Point(BN254.G1Point memory pk) internal pure returns (bytes32) {
-        return keccak256(abi.encodePacked(pk.X, pk.Y));
+    function hashG1Point(BN254.G1Point memory pk) internal pure returns (bytes32 hashedG1) {
+        assembly {
+            mstore(0, mload(pk))
+            mstore(0x20, mload(add(0x20, pk)))
+            hashedG1 := keccak256(0, 0x40)
+        }
```
0x0aa0, line 278: very nice

0x0aa0 (approved):
lgtm

Issue #103, opened 2023-12-13
Solidity Compilation Error on Importing BLSApkRegistry: Base Must Precede Derived Contract


ny-alt:
**Describe the bug**

An error occurs during the compilation of [Sample.sol](https://github.com/ny-alt/sample-fe3ae2/blob/master/src/Sample.sol), specifically after importing BLSApkRegistry.

Sample.sol
```
// SPDX-License-Identifier: UNLICENSED
pragma solidity =0.8.12;

import {BLSApkRegistry} from "eigenlayer-middleware/src/BLSApkRegistry.sol";
``` 

Compilation Error
```
Compiler run failed:
Error (2449): Definition of base has to precede definition of derived contract
  --> src/interfaces/IStakeRegistry.sol:13:29:
   |
13 | interface IStakeRegistry is IRegistry {
   |                             ^^^^^^^^^
```

Detailed error logs and context can be found in the GitHub Actions workflow at [this link](https://github.com/ny-alt/sample-fe3ae2/actions/runs/7196849650/job/19602614913).

**To Reproduce**
1. git clone https://github.com/ny-alt/sample-fe3ae2.git
2. forge compile

**Expected behavior**
The expected behavior is that the Solidity project should compile without errors.

ChaoticWalrus:
I experienced this error recently, and am trying to sort out the root cause, because it appeared while making unrelated changes.

@ny-alt a workaround for you, that has worked for me repeatedly:

1. run `forge clean`
2. immediately followed by whatever other command you're trying to run (`forge b`, `forge t`, etc.)

Obviously this isn't ideal, but let me know if it doesn't fix the problem for you

ChaoticWalrus:
This should have been fixed in https://github.com/Layr-Labs/eigenlayer-middleware/pull/108
Please reopen this issue if that PR doesn't sort things out for you.

PR #108, opened 2023-12-13
fix: remove circular dependency

wadealexc:


ChaoticWalrus (approved):
Looks appears to fixes the pesky, flaky compilation errors we were seeing internally + were reported externally in https://github.com/Layr-Labs/eigenlayer-middleware/issues/103
It's a bit hard to tell due to their weird, unstable nature, but the root cause certainly seemed to be a circular dependency which this PR addresses.

Changes are more minimal than I expected and all look fine.

Approving and merging now.

PR #109, opened 2023-12-14
feat: concentrate eigenlayer avs interactions into service manager

ChaoticWalrus:
# Summary
Reintroduce an IServiceManager interface and ServiceManagerBase, but including only on the very minimal interactions we are requiring from AVSs (the other one was very slashing-focused, so there really isn’t overlap between the old interface and this new one)

## Why?
We clearly aren’t going to be able to jam that much more into the RegistryCoordinator, and setting EigenLayer up to have to worry about multiple AVS contracts is a recipe for headaches down the line.
We know we’ll be adding more interactions where AVSs push data to EigenLayer in the future, and it seems highly unlikely we can fit these into the RegistryCoordinator, or that these would be the right place even if we could.
I also feel like this is better separation of concerns. Hopefully AVS developers have less to worry about here — if they don’t want to change the Registries we have, they can just worry about customizing the ServiceManager / adding more functionality there down the line, rather than needing to fit things into the RegistryCoordinator.
We are making the clear decision of “one contract per AVS for pushing info to EigenLayer” which we had tried to stick with before for good reason. It places a bit of a constraint on AVS architecture (one kind of central router for calls to EigenLayer) but I think less than saying “you need to follow this registry structure” which it feels like we’re implicitly heading into without doing something like this.

## Open Questions
Do we need the IServiceManager interface at all?
~~If so, should we perhaps add the `setMetadataURI` function to it?~~ (added in https://github.com/Layr-Labs/eigenlayer-middleware/pull/109/commits/a8008a07656c082ab9e2df0d8c9543b3d0734392)

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -3,11 +3,13 @@ pragma solidity =0.8.12;
 
 import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 
+import {BitmapUtils} from "src/libraries/BitmapUtils.sol"; 
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
 
 import {IServiceManager} from "src/interfaces/IServiceManager.sol";
 import {IRegistryCoordinator} from "src/interfaces/IRegistryCoordinator.sol";
+import {IStakeRegistry} from "src/interfaces/IStakeRegistry.sol";
 
 /**
  * @title Minimal implementation of a ServiceManager-type contract.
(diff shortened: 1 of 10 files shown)


src/ServiceManagerBase.sol
```diff
@@ -0,0 +1,72 @@
...
+    function initialize(address initialOwner) external initializer {
+        _transferOwnership(initialOwner);
+    }
+
+    /**
+     * @notice Sets the metadata URI for the AVS
+     * @param _metadataURI is the metadata URI for the AVS
+     * @dev only callable by the owner
+     */
+    function setMetadataURI(string memory _metadataURI) public virtual onlyOwner {
```
samlaf, line 49: why is this not in the IServiceManager interface?

ChaoticWalrus: I asked that in the PR description. Lemme just add it real quick; I think if we have the IServiceManager interface at all, this makes sense to be in it.

ChaoticWalrus: OK, added it in https://github.com/Layr-Labs/eigenlayer-middleware/pull/109/commits/a8008a07656c082ab9e2df0d8c9543b3d0734392

wadealexc (approved):
I'm fine with this. Makes sense to avoid locking an AVS into our registry structure. I like the minimal interface.

:+1:

ypatil12 (approved):
LGTM - I can update the tests in my PR

PR #110, opened 2023-12-14
Feat: Add Operator Strategy Indexing to Service Manager

ypatil12:
We need to index operators restaked on each AVS to support our user stories for [V1 AVS Marketplace](https://docs.google.com/document/d/17PStsH5qXFNWUzjzr-vyzASy_6bjMd_tuf234bkejkg/edit).

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -3,11 +3,13 @@ pragma solidity =0.8.12;
 
 import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 
+import {BitmapUtils} from "src/libraries/BitmapUtils.sol"; 
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
 
 import {IServiceManager} from "src/interfaces/IServiceManager.sol";
 import {IRegistryCoordinator} from "src/interfaces/IRegistryCoordinator.sol";
+import {IStakeRegistry} from "src/interfaces/IStakeRegistry.sol";
 
 /**
  * @title Minimal implementation of a ServiceManager-type contract.
@@ -15,8 +17,11 @@ import {IRegistryCoordinator} from "src/interfaces/IRegistryCoordinator.sol";
  * @author Layr Labs, Inc.
  */
 contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
-    address immutable registryCoordinator;
+    using BitmapUtils for *;
+
+    IRegistryCoordinator immutable registryCoordinator;
     IDelegationManager immutable delegationManager;
+    IStakeRegistry immutable stakeRegistry;
 
     /// @notice when applied to a function, only allows the RegistryCoordinator to call it
     modifier onlyRegistryCoordinator() {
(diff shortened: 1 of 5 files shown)


ChaoticWalrus (commented):
Looks fine to me. Want to get some more discussion on this (probably offline) but I think this is a reasonable proposal on the face of it.
Not noticing any issues with the code from a quick review.

src/ServiceManagerBase.sol
```diff
@@ -69,4 +76,40 @@ contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
...
+
+        if (operatorBitmap == 0 || registryCoordinator.quorumCount() == 0) {
+            return new address[](0);
+        }
+
+        // Get number of strategies for each quorum in operator bitmap
+        bytes memory operatorRestakedQuorums = BitmapUtils.bitmapToBytesArray(operatorBitmap);
+        uint256 strategyCount;
+        for(uint256 i = 0; i < operatorRestakedQuorums.length; i++) {
+            strategyCount += stakeRegistry.strategyParamsLength(uint8(operatorRestakedQuorums[i]));
```
gpsanant, line 99: will be an overestimate if there are duplicates

src/ServiceManagerBase.sol
```diff
@@ -3,20 +3,25 @@ pragma solidity =0.8.12;
...
+import {IStakeRegistry} from "src/interfaces/IStakeRegistry.sol";
 
 /**
  * @title Minimal implementation of a ServiceManager-type contract.
  * This contract can inherited from or simply used as a point-of-reference.
  * @author Layr Labs, Inc.
  */
 contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
-    address immutable registryCoordinator;
+    using BitmapUtils for *;
```
gpsanant, line 20: nit: can we change this

src/ServiceManagerBase.sol
```diff
@@ -69,4 +76,40 @@ contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
...
     }
+
+    /**
+     * @notice Returns the list of strategies that the operator has potentially restaked on the AVS
+     * @param operator The address of the operator to get restaked strategies for
+     * @dev This function is intended to be called off-chain
+     * @dev No guarantee is made on whether the operator has shares for a strategy in a quorum or uniqueness 
+     *      of each element in the returned array. The off-chain service should do that validation separately
+     */
+    function getOperatorRestakedStrategies(address operator) external view returns (address[] memory) {
```
gpsanant, line 87: we need to set a gas limit which we will call this with to make sure we're clear about how errors are handled

src/ServiceManagerBase.sol
ypatil12: Yup, noted in dev comment

src/ServiceManagerBase.sol
```diff
@@ -69,4 +76,40 @@ contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
...
+        bytes32 operatorId = registryCoordinator.getOperatorId(operator);
+        uint192 operatorBitmap = registryCoordinator.getCurrentQuorumBitmap(operatorId);
+
+        if (operatorBitmap == 0 || registryCoordinator.quorumCount() == 0) {
+            return new address[](0);
+        }
```
wadealexc, line 93: Second part of this bool should result in the same outcome - maybe it's not needed?

src/ServiceManagerBase.sol
wadealexc: I don't think this is an estimate so much as a specific count. If there are duplicates, the return array will have duplicates

I don't think that's a problem given it's consumed by offchain code though

src/ServiceManagerBase.sol
ypatil12: updated since we're calling into library directly

ChaoticWalrus:
I'm in approval as well. Merging this and the PR that it's targeting after that.

PR #113, opened 2023-12-15
feat: Add restakeable strategies for indexing

ypatil12:
We need to also index all strategies that are restakeable, since the UI displays 0 TVL for strategies not restaked on yet.


wadealexc (approved):
Clean, LGTM.

ChaoticWalrus (approved):
LGTM, keeping it surgical 👍

PR #116, opened 2023-12-18
docs: update reg coord to include pubkey registration and service man…

wadealexc:
…ager usage


docs/BLSSignatureChecker.md
```diff
@@ -1 +1,196 @@
...
+struct QuorumStakeTotals {
+    uint96[] signedStakeForQuorum;
+    uint96[] totalStakeForQuorum;
+}
+```
+
+The goal of this method is to allow an AVS to validate a BLS signature formed from the aggregate pubkey ("apk") of Operators registered in one or more quorums at some `referenceBlockNumber`.
+
+Some notes on method parameters:
+* `referenceBlockNumber` is the reason each registry contract keeps historical states: so that lookups can be performed on each registry's info at a particular block. This is important because Operators may sign some data on behalf of an AVS, then deregister from one or more of the AVS's quorums. Historical states allow signature validation to be performed against a "fixed point" in AVS/quorum history.
```
ChaoticWalrus, line 65: this also helps deal with race conditions. if all you keep track of is the current APK, then when a new operator registered, it would cause all "in-flight" confirmations to revert.

ChaoticWalrus: ah, on re-reading, maybe that's what you're already saying 😄

docs/BLSSignatureChecker.md
```diff
@@ -1 +1,196 @@
...
+* `params` contains both a signature from all signing Operators, as well as several fields that identify registered, non-signing Operators. While non-signing Operators haven't contributed to the signature, but need to be accounted for because, as Operators registered for one or more signing quorums, their public keys are included in that quorum's apk. Essentially, in order to validate the signature, nonsigners' public keys need to be subtracted out from the total apk to derive the apk that actually signed the message.
+
+This method performs the following steps. Note that each step involves lookups of historical state from `referenceBlockNumber`, but the writing in this section will use the present tense because adding "at the `referenceBlockNumber`" everywhere gets confusing. Steps follow:
+1. Calculate a total apk for all nonsigners, multiplying each nonsigner's pubkey by the number of `quorumNumbers` they are registered for.
+2. Negate the calculated total apk to "remove" the nonsigners' public keys.
```
ChaoticWalrus: might want to call this the "total _nonsigner_ apk" for the sake of clarity?

ChaoticWalrus: also feels like there's maybe a missing operation of:
2.5 Add the negated nonsigner APK to the quorum's APK, to get the APK for all signing operators of the quorum

perhaps this is more confusing, but that's technically what's happening.

docs/BLSSignatureChecker.md
```diff
@@ -1 +1,196 @@
...
+3. For each quorum:
+    * Add the quorum's total apk to the total apk being calculated. Because the negation of nonsigner pubkeys are already in the total apk, this effectively subtracts nonsigner pubkeys from the total apk (leaving only signing pubkeys)
```
ChaoticWalrus: likewise, I think "total apk" -> "signing apk" could help with clarity/precision.
So I would revise this line to something like
```
    * Add the quorum's signing apk to the total (signing) apk being calculated
```
at least if you make the above addition to clarify how this signing apk is calculated

ChaoticWalrus (approved):
LGTM 👍 
Primarily had some minor clarity suggestions on the APK stuff.

docs/BLSSignatureChecker.md
wadealexc: after i refactored we only do that "step 2.5" a single time - the total nonsigner apk is negated once, then each quorum is re-added.

i rewrote a bit to make it more clear; commit pending.

docs/BLSSignatureChecker.md
wadealexc: [20583ba](https://github.com/Layr-Labs/eigenlayer-middleware/pull/116/commits/20583bab55930680795024e599a70ec71a400325)

docs/BLSSignatureChecker.md
wadealexc: [20583ba](https://github.com/Layr-Labs/eigenlayer-middleware/pull/116/commits/20583bab55930680795024e599a70ec71a400325)

PR #117, opened 2023-12-18
chore: make immutable variables in ServiceManagerBase explicitly inte…

ChaoticWalrus:
…rnal

I believe this should help resolve some inheritance issues we were facing. Commit also makes the `initialize` function public and virtual for easier overriding


samlaf (approved):
LGTM, thanks Jeff :)

PR #120, opened 2023-12-21
Test/bls operator state ret unit tests

gpsanant:


test/tree/OperatorStateRetrieverUnit.tree
```diff
@@ -0,0 +1,35 @@
+.
+└── OperatorStateRetriever tree (*** denotes that integration tests are needed to validate path)
+    ├── when getOperatorState(IRegistryCoordinator,bytes32,uint32) is called
+    │   ├── when the given operator has never registered with the given registryCoordinator (their quorumBitmapHistory has length 0)
+    │   │   └── it should revert with "RegistryCoordinator.getQuorumBitmapIndicesAtBlockNumber: operator has no bitmap history"
```
8sunyuan, line 5: This and line 7 should be
"RegistryCoordinator.getQuorumBitmapIndicesAtBlockNumber: operatorId has no quorumBitmaps at blockNumber"

Just writing "it should revert" also works too

test/tree/OperatorStateRetrieverUnit.tree
```diff
@@ -0,0 +1,35 @@
+.
+└── OperatorStateRetriever tree (*** denotes that integration tests are needed to validate path)
+    ├── when getOperatorState(IRegistryCoordinator,bytes32,uint32) is called
+    │   ├── when the given operator has never registered with the given registryCoordinator (their quorumBitmapHistory has length 0)
```
8sunyuan, line 4: Perhaps `_operatorBitmapHistory` is more explicit and accurate here instead of `quorumBitmapHistory`

test/tree/OperatorStateRetrieverUnit.tree
```diff
@@ -0,0 +1,35 @@
...
+    │   │   └── it should revert with "RegistryCoordinator.getQuorumBitmapIndicesAtBlockNumber: operator has no bitmap history"
+    │   └── when the given operator has registered for the first time with the given registryCoordinator
+    │       ├── after the given blockNumber
+    │       │   └── it should revert with "RegistryCoordinator.getQuorumBitmapIndicesAtBlockNumber: operator has no bitmap history at blockNumber"
+    │       └── before the given blockNumber
+    │           ├── and was not registered for any quorums at the given blockNumber
```
8sunyuan, line 10: I believe an invariant of the RegistryCoordinator is that a registered operator must have a nonzero bitmap.
So if the operator hasn't registered for any quorums i.e zero bitmap, they would be deregistered.

test/tree/OperatorStateRetrieverUnit.tree
```diff
@@ -0,0 +1,35 @@
...
+    │           │   └── it should return (0, [])
+    │           └── and was registered for a set of quorums at the given blockNumber
+    │               └── it should return their quorum bitmap as the first argument and a list of the list of operators (by id and stake) ordered by index for each quorum in the set ascending by quorumNumber
+    ├── when getOperatorState(IRegistryCoordinator,bytes memory,uint32) is called
+    │   ├── when at any of the given quorumNumbers are not created at the time of call
+    │   │   └── it should revert with "IndexRegistry._operatorCountAtBlockNumber: quorum does not exist"
+    │   ├── when at any of the given quorumNumbers are not created before or at the given blockNumber
+    │   │   └── it should revert with "IndexRegistry._operatorCountAtBlockNumber: quorum did not exist at given block number"
+    │   └── when the given quorumNumbers are created on the given registry registryCoordinator
+    │       └── it should return their quorum bitmap as the first argument and a list of the list of operators (by id and stake) ordered by index for each quorum in the set ascending by quorumNumber
```
8sunyuan, line 20: Doesn't return a bitmap like the above function

src/RegistryCoordinator.sol
```diff
@@ -695,6 +695,13 @@ contract RegistryCoordinator is
         }
     }
 
+    /**
+     * @notice Returns true iff all of the given quorum exists
+     */
+    function _quorumExists(uint8 quorumNumber) internal view returns (bool) {
```
8sunyuan, line 701: If we use this, might be cleaner to just scrap the `modifier _quorumExists` and use the function version.
The modifier is only being called in `setOperatorSetParams`

gpsanant:
Closed in favor of https://github.com/Layr-Labs/eigenlayer-middleware/pull/140

PR #121, opened 2023-12-21
feat: registry coordinator unit test improvements

ChaoticWalrus:
Initially marking as a draft since this work is definitely not complete.


ChaoticWalrus:
This is looking better now.
TODOs (editing as I add coverage):
testing for ~~`updateOperators`, `updateOperatorsForQuorum`, `getQuorumBitmapIndicesAtBlockNumber`, `getQuorumBitmapAtBlockNumberByIndex`,`_registerOperator`, `_deregisterOperator`,~~ `_updateOperator`, ~~and `_updateOperatorBitmap`~~

ChaoticWalrus:
I'm unsure if adding coverage for `_updateOperator` to these unit tests makes sense, as it is mostly a call to the StakeRegistry / feels a bit more apt as an "integration test".
Going to give a little more thought to the testing of the view functions.
Otherwise this PR is now looking good, I think.

ChaoticWalrus:
I believe this PR is now ready for review.
A few things of note:

1. this PR mildly modifies the behavior of the `getQuorumBitmapIndicesAtBlockNumber`, making it revert any of the operators had not yet registered at the `blockNumber` in question -- see new code here https://github.com/Layr-Labs/eigenlayer-middleware/blob/4e70e8126a1259ac0172c456512117cd25db6c79/src/RegistryCoordinator.sol#L761-L801
2. I didn't add coverage for the internal `_updateOperator` function (see my reasoning here https://github.com/Layr-Labs/eigenlayer-middleware/pull/121#issuecomment-1877915543)
3. For (1) above, it seems we _could_ remove the require statement in these lines: https://github.com/Layr-Labs/eigenlayer-middleware/blob/4e70e8126a1259ac0172c456512117cd25db6c79/src/RegistryCoordinator.sol#L789-L794 as it doesn't seem our storage model will ever trigger this particular revert.
However, I kind of like leaving it in as a sanity check. I believe the same applies to the equivalent code in the StakeRegistry (https://github.com/layr-labs/eigenlayer-middleware/blob/98f884454d9e9de1e344bb6fba9a2cd3915e5b57/src/StakeRegistry.sol#L297-L299).

src/RegistryCoordinator.sol
```diff
@@ -758,33 +758,53 @@ contract RegistryCoordinator is
...
-     * @notice Returns the quorum bitmap for the given `operatorId` at the given `blockNumber` via the `index`
-     * @dev reverts if `index` is incorrect 
+     * @notice Returns the index of the quorumBitmap for the provided `operatorId` at the given `blockNumber`
+     * @dev Reverts if the operator had not yet (ever) registered at `blockNumber`
+     * @dev This function is designed to find proper inputs to the `getQuorumBitmapAtBlockNumberByIndex` function
+     */
+    function _getQuorumBitmapIndexAtBlockNumber(
+        uint32 blockNumber, 
+        bytes32 operatorId
+    ) internal view returns (uint32 index) {
```
wadealexc: We have a section for internal functions, this should be in there. This bottom section is for external/public getters (though I see the comment only says "view functions", maybe that should be clarified)

src/RegistryCoordinator.sol
```diff
@@ -758,33 +758,53 @@ contract RegistryCoordinator is
...
+        uint256 length = _operatorBitmapHistory[operatorId].length;
+        for (uint256 i = 0; i < length; i++) {
+            if (_operatorBitmapHistory[operatorId][length - i - 1].updateBlockNumber <= blockNumber) {
+                uint32 nextUpdateBlockNumber = 
+                    _operatorBitmapHistory[operatorId][length - i - 1].nextUpdateBlockNumber;
+                require(
+                    nextUpdateBlockNumber == 0 || nextUpdateBlockNumber > blockNumber,
+                    "RegistryCoordinator.getQuorumBitmapIndexAtBlockNumber: operatorId has no bitmap update at blockNumber"
+                );
```
wadealexc: I'm less a fan of leaving this in, for a few reasons. 

First, if we hit this, it means the contract's storage is corrupted and there's a bug outside this method.

IMO contracts should keep assertions like this to a minimum, primarily to cover important state changes. This is a getter used by offchain code, so I feel like it only confuses (I mean, we know this code and we had a hard time figuring out why it was here)

Second - if we really want to keep it, let's at least change the error message - it's wildly inaccurate and will lead to further confusion. Maybe something like `RegCoord.methodName: history is corrupted`?

Third - this method is really clean and straightforward without these extra lines of code. I like that, because currently the "state history" model our code uses is really confusing, and being able to look at the getters our offchain code is consuming makes things make more sense - unless they're bogged down with stuff like this

src/RegistryCoordinator.sol
```diff
@@ -758,33 +758,53 @@ contract RegistryCoordinator is
...
+                return uint32(length - i - 1);
```
wadealexc: Since the return param is named, maybe we should assign it and replace the two `length - i - 1` parts with `index`?

I feel like it makes the "index" concept a bit more tangible.

test/unit/RegistryCoordinatorUnit.t.sol
```diff
@@ -853,157 +1493,285 @@ contract RegistryCoordinatorUnit is MockAVSDeployer {
...
 
-        cheats.expectEmit(true, true, true, true, address(blsApkRegistry));
-        emit OperatorRemovedFromQuorums(defaultOperator, quorumNumbers);
+        address[] memory operatorsToUpdate = new address[](1);
+        operatorsToUpdate[0] = defaultOperator;
 
-        cheats.expectEmit(true, true, true, true, address(stakeRegistry));
-        emit OperatorStakeUpdate(defaultOperatorId, uint8(quorumNumbers[0]), 0);
+        // TODO: any additional checks here? mostly this just calls the StakeRegistry, so more appropriate for an integration test
+        registryCoordinator.updateOperators(operatorsToUpdate);
```
wadealexc: Probably worth having a test that checks that if the StakeRegistry returns quorums to remove from the operator, that they're removed?

wadealexc (approved):
Looks clean - a few small requests, but I'm gonna go ahead and approve. Just make sure the latest m2-mainnet is merged in, since it looks like this is a merge rather than a rebase :)

src/RegistryCoordinator.sol
ChaoticWalrus: fair enough. I pointed out that it could be removed to elicit some feedback. I don't feel strongly either way, but you've convinced me to remove it.

src/RegistryCoordinator.sol
ChaoticWalrus: ah, my bad, just forgot to put it in the right section, will move it.

src/RegistryCoordinator.sol
ChaoticWalrus: good suggestion, I'll make the change

test/unit/RegistryCoordinatorUnit.t.sol
ChaoticWalrus: hmm yeah, I wasn't sure if this made more sense in "integration" tests. right now these tests use a ~~mock~~ harnessed StakeRegistry; I'll see if I can cook a good test up here.

wadealexc: It's definitely getting hit in integration tests, so up to you if you want to leave it out!

src/RegistryCoordinator.sol
ChaoticWalrus: done here https://github.com/Layr-Labs/eigenlayer-middleware/pull/121/commits/ab44a8239660a5fb469922e76b1b5921e0d1b5e8

src/RegistryCoordinator.sol
ChaoticWalrus: code removed here https://github.com/Layr-Labs/eigenlayer-middleware/pull/121/commits/8cd21f4d8754453d5f90948b9be74fe2885d8230

src/RegistryCoordinator.sol
ChaoticWalrus: made this change in this commit https://github.com/Layr-Labs/eigenlayer-middleware/pull/121/commits/8cd21f4d8754453d5f90948b9be74fe2885d8230

test/unit/RegistryCoordinatorUnit.t.sol
ChaoticWalrus: test added in https://github.com/Layr-Labs/eigenlayer-middleware/pull/121/commits/1f1a840d0b83c27e71a940f9e64ae4bb0a1863ba
I think this test worked as a minimal version, but I added a more thorough fuzzed one.

PR #122, opened 2023-12-27
feat: public immutable vars in ServiceManagerBase

ghost:
This PR updates the visibility of three internal state variables (_registryCoordinator, _delegationManager, and _stakeRegistry) to public. Additionally, immutable getters are introduced for these variables.

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -3,7 +3,7 @@ pragma solidity =0.8.12;
 
 import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 
-import {BitmapUtils} from "./libraries/BitmapUtils.sol"; 
+import {BitmapUtils} from "./libraries/BitmapUtils.sol";
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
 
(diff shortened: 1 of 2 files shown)


wadealexc:
Thanks for the PR. We just updated these to be internal - do you need them to be public for some reason?

ny-alt:
Yes, it is for the getters.

ghost:
> Thanks for the PR. We just updated these to be internal - do you need them to be public for some reason?

Yes, it's for the getters in my case.

wadealexc:
i see - i think `ServiceManagerBase` is meant to be something you inherit from. Maybe you could create a `ServiceManager is ServiceManagerBase` for your usecase, and expose the variables there?

Maybe we should mark `ServiceManagerBase` as `abstract` :sweat:

ghost:
I understand that ServiceManagerBase is meant for inheritance. 

What are the cons of making those variables as `public`?

wadealexc:
I think someone on our team was making a service manager contract that also inherits from `BLSSignatureChecker`, and there were some issues with shadowing/overloading.

wadealexc:
Sorry, I'm not 100% sure -- this happened while I was on vacation :D

ghost:
I see. But It no longer inherits `BLSSignatureChecker`.

wadealexc:
Correct - they wanted to make a contract like this:

`contract ServiceManager is ServiceManagerBase, BLSSignatureChecker`

wadealexc:
Going to close this for now, but thank you for the PR.

Also if you want to chat about anything else on our Discord (rather than over PR comments), we have a `#talk-smart-contracts` channel that's perfect for this! https://discord.gg/wfG9Kr8h

PR #123, opened 2023-12-28
feat: experimental ecdsa registries

ChaoticWalrus:
Extremely rough draft contracts that I threw together in ~90 minutes.

Do NOT use these contracts without some kind of review.

Comments in the contracts are definitely a bit of a mess -- see the commit messages for a brief summary of work here.

### DIFF
--- a/src/ECDSARegistryCoordinator.sol
+++ b/src/ECDSARegistryCoordinator.sol
@@ -1,0 +1,867 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import {Initializable} from "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
+
+import {EIP1271SignatureUtils} from "eigenlayer-contracts/src/contracts/libraries/EIP1271SignatureUtils.sol";
+import {IPauserRegistry} from "eigenlayer-contracts/src/contracts/interfaces/IPauserRegistry.sol";
+import {Pausable} from "eigenlayer-contracts/src/contracts/permissions/Pausable.sol";
+
+import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
+import {ISocketUpdater} from "./interfaces/ISocketUpdater.sol";
+import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
+import {IServiceManager} from "./interfaces/IServiceManager.sol";
+
+import {BitmapUtils} from "./libraries/BitmapUtils.sol";
+
+/**
+ * @title A `RegistryCoordinator` that has three registries:
+ *      1) a `StakeRegistry` that keeps track of operators' stakes
+ *      2) a `BLSApkRegistry` that keeps track of operators' BLS public keys and aggregate BLS public keys for each quorum
+ *      3) an `IndexRegistry` that keeps track of an ordered list of operators for each quorum
+ * 
+ * @author Layr Labs, Inc.
+ */
+contract ECDSARegistryCoordinator is 
+    EIP712, 
+    Initializable, 
+    Pausable,
+    OwnableUpgradeable,
+    ISocketUpdater, 
+    ISignatureUtils
+{
+    using BitmapUtils for *;
+
+    // EVENTS
+
+    /// Emits when an operator is registered
+    event OperatorRegistered(address indexed operator, bytes32 indexed operatorId);
+
+    /// Emits when an operator is deregistered
+    event OperatorDeregistered(address indexed operator, bytes32 indexed operatorId);
+
+    event OperatorSetParamsUpdated(uint8 indexed quorumNumber, OperatorSetParam operatorSetParams);
+
+    event ChurnApproverUpdated(address prevChurnApprover, address newChurnApprover);
+
+    event EjectorUpdated(address prevEjector, address newEjector);
+
+    /// @notice emitted when all the operators for a quorum are updated at once
+    event QuorumBlockNumberUpdated(uint8 indexed quorumNumber, uint256 blocknumber);
+
+    // DATA STRUCTURES
+    enum OperatorStatus
+    {
+        // default is NEVER_REGISTERED
+        NEVER_REGISTERED,
+        REGISTERED,
+        DEREGISTERED
+    }
+
+    // STRUCTS
+
+    /**
+     * @notice Data structure for storing info on operators
+     */
+    struct OperatorInfo {
+        // the id of the operator, which is likely the keccak256 hash of the operator's public key if using BLSRegistry
+        bytes32 operatorId;
+        // indicates whether the operator is actively registered for serving the middleware or not
+        OperatorStatus status;
+    }
+
+    /**
+     * @notice Data structure for storing operator set params for a given quorum. Specifically the 
+     * `maxOperatorCount` is the maximum number of operators that can be registered for the quorum,
+     * `kickBIPsOfOperatorStake` is the basis points of a new operator needs to have of an operator they are trying to kick from the quorum,
+     * and `kickBIPsOfTotalStake` is the basis points of the total stake of the quorum that an operator needs to be below to be kicked.
+     */ 
+     struct OperatorSetParam {
+        uint32 maxOperatorCount;
+        uint16 kickBIPsOfOperatorStake;
+        uint16 kickBIPsOfTotalStake;
+    }
+
+    /**
+     * @notice Data structure for the parameters needed to kick an operator from a quorum with number `quorumNumber`, used during registration churn.
+     * `operator` is the address of the operator to kick
+     */
+    struct OperatorKickParam {
+        uint8 quorumNumber;
+        address operator;
+    }
+
+    // TODO: doument
+    struct ECDSAPubkeyRegistrationParams {
+        address signingAddress;
+        SignatureWithSaltAndExpiry signatureAndExpiry;
+    }
+
+    /// @notice The EIP-712 typehash for the `DelegationApproval` struct used by the contract
+    bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
+        keccak256("OperatorChurnApproval(bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams)OperatorKickParam(address operator,bytes32[] operatorIdsToSwap)");
+    /// @notice The EIP-712 typehash used for registering BLS public keys
+    bytes32 public constant PUBKEY_REGISTRATION_TYPEHASH = keccak256("ECDSAPubkeyRegistration(address operator)");
+    /// @notice The maximum value of a quorum bitmap
+    uint256 internal constant MAX_QUORUM_BITMAP = type(uint256).max;
+    /// @notice The basis point denominator
+    uint16 internal constant BIPS_DENOMINATOR = 10000;
+    /// @notice Index for flag that pauses operator registration
+    uint8 internal constant PAUSED_REGISTER_OPERATOR = 0;
+    /// @notice Index for flag that pauses operator deregistration
+    uint8 internal constant PAUSED_DEREGISTER_OPERATOR = 1;
+    /// @notice Index for flag pausing operator stake updates
+    uint8 internal constant PAUSED_UPDATE_OPERATOR = 2;
+    /// @notice The maximum number of quorums this contract supports
+    uint8 internal constant MAX_QUORUM_COUNT = 192;
+
+    /// @notice the ServiceManager for this AVS, which forwards calls onto EigenLayer's core contracts
+    IServiceManager public immutable serviceManager;
+    /// @notice the Stake Registry contract that will keep track of operators' stakes
+    IStakeRegistry public immutable stakeRegistry;
+
+    /// @notice the current number of quorums supported by the registry coordinator
+    uint8 public quorumCount;
+    /// @notice maps quorum number => operator cap and kick params
+    mapping(uint8 => OperatorSetParam) internal _quorumParams;
+    /// @notice maps operator id => current quorums they are registered for
+    mapping(bytes32 => uint256) public operatorBitmap;
+    /// @notice maps operator address => operator id and status
+    mapping(address => OperatorInfo) internal _operatorInfo;
+    /// @notice whether the salt has been used for an operator churn approval
+    mapping(bytes32 => bool) public isChurnApproverSaltUsed;
+    /// @notice mapping from quorum number to the latest block that all quorums were updated all at once
+    mapping(uint8 => uint256) public quorumUpdateBlockNumber;
+
+    // TODO: document
+    mapping(bytes32 => address) operatorIdToOperator;
+    mapping(uint8 => uint32) public totalOperatorsForQuorum;
+
+    /// @notice the dynamic-length array of the registries this coordinator is coordinating
+    address[] public registries;
+    /// @notice the address of the entity allowed to sign off on operators getting kicked out of the AVS during registration
+    address public churnApprover;
+    /// @notice the address of the entity allowed to eject operators from the AVS
+    address public ejector;
+
+    modifier onlyEjector {
+        require(msg.sender == ejector, "RegistryCoordinator.onlyEjector: caller is not the ejector");
+        _;
+    }
+
+    modifier quorumExists(uint8 quorumNumber) {
+        require(
+            quorumNumber < quorumCount, 
+            "RegistryCoordinator.quorumExists: quorum does not exist"
+        );
+        _;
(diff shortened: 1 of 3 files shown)


src/ECDSASignatureChecker.sol
```diff
@@ -0,0 +1,170 @@
...
+
+import {BitmapUtils} from "./libraries/BitmapUtils.sol";
+
+/**
+ * @title Used for checking BLS aggregate signatures from the operators of a `BLSRegistry`.
+ * @author Layr Labs, Inc.
+ * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
+ * @notice This is the contract for checking the validity of aggregate operator signatures.
+ */
+contract BLSSignatureChecker {
```
samlaf, line 16: Update contract name. also ctrl-f "aggregate" and "bls" and remove any reference.

samlaf (commented):
Seems like an awesome first start, thanks for this.
One big area of complexity that would be nice to get rid of is the churner, but unsure if there are alternatives that have been considered that could work better with this simpler registry?

src/ECDSARegistryCoordinator.sol
```diff
@@ -0,0 +1,867 @@
...
+    uint8 internal constant PAUSED_DEREGISTER_OPERATOR = 1;
+    /// @notice Index for flag pausing operator stake updates
+    uint8 internal constant PAUSED_UPDATE_OPERATOR = 2;
+    /// @notice The maximum number of quorums this contract supports
+    uint8 internal constant MAX_QUORUM_COUNT = 192;
+
+    /// @notice the ServiceManager for this AVS, which forwards calls onto EigenLayer's core contracts
+    IServiceManager public immutable serviceManager;
+    /// @notice the Stake Registry contract that will keep track of operators' stakes
+    IStakeRegistry public immutable stakeRegistry;
```
samlaf, line 123: ECDSAStakeRegistry does not implement IStakeRegistry. Should this be ECDSAStakeRegistry instead of IStakeRegistry? Or should ECDSAStakeRegistry implement IStakeRegistry?

src/ECDSASignatureChecker.sol
ChaoticWalrus: Yes this would need a lot of cleanup w/r/t naming and comments, in particular

src/ECDSARegistryCoordinator.sol
ChaoticWalrus: I was super liberal with removing inheritance for the sake of expediency here.
It's very possible the integration between these contracts straight-up doesn't work rn, but it would be pretty quick to check if the ECDSAStakeRegistry implements all the functions we're calling from the IStakeRegistry interface (I think it does?)
If you want this developed further, then the path forward isn't immediately clear to me.
Perhaps making a new `IECDSAStakeRegistry` interface or something would make sense, or these changes could inform _reducing_ the scope of the `IStakeRegistry` interface so both can fit into it, and then the more specific `IECDSAStakeRegistry` and `IBLSStakeRegistry` interfaces (if these need to exist at all!) can both inherit from this minimal interface.

src/ECDSASignatureChecker.sol
```diff
@@ -0,0 +1,170 @@
...
+     * is correct, i.e., ensure that the stake returned from the specified block number is recent enough and that the stake is either the most recent update
+     * for the total stake (of the operator) or latest before the referenceBlockNumber.
+     * @param msgHash is the hash being signed
+     * @param quorumNumbers is the bytes array of quorum numbers that are being signed for
+     * @param signerIds TODO: document
+     * @param signatures TODO: document
+     * @return quorumStakeTotals is the struct containing the total and signed stake for each quorum
+     * @return signatoryRecordHash is the hash of the signatory record, which is used for fraud proofs
+     */
+    function checkSignatures(
```
samlaf, line 86: Implementing the ecdsa version of inc-sq using these contracts and just realized this doesn't have a `referenceBlockNumber` like the bls version. Does this mean that the ecdsa stakeregistry doesn't track weight histories and can only validate against current weights?

src/ECDSASignatureChecker.sol
```diff
@@ -0,0 +1,170 @@
...
+        bytes32 msgHash, 
+        bytes calldata quorumNumbers,
+        bytes32[] memory signerIds,
```
samlaf, line 89: should we be calling `signerIds` `signerAddresses` instead? 
They are verified in
```
checkSignature_EIP1271(address signer, bytes32 digestHash, bytes memory signature)
```
and passed as the signer (an address)

samlaf (commented):
Also realizing we might need an ECDSAOperatorStateRetriever?

src/ECDSASignatureChecker.sol
ChaoticWalrus: in this quick + hacked-together version I completely removed histories, so yes, you are exactly correct

src/ECDSASignatureChecker.sol
ChaoticWalrus: fine by me!

ChaoticWalrus:
> Also realizing we might need an ECDSAOperatorStateRetriever?

yes, this would also likely be a practical or even necessary addition.

src/ECDSARegistryCoordinator.sol
```diff
@@ -0,0 +1,867 @@
...
+    }
+
+    // STRUCTS
+
+    /**
+     * @notice Data structure for storing info on operators
+     */
+    struct OperatorInfo {
+        // the id of the operator, which is likely the keccak256 hash of the operator's public key if using BLSRegistry
+        bytes32 operatorId;
```
samlaf, line 70: Still a bit confused as to whether this should be stored as a bytes32 or an address. 
```
which is likely the keccak256 hash of the operator's public key if using BLSRegistry
```
shows that this was copied from the BLSRegistry. Pretty convinced this should be address here right? (hash of ecdsa pubkey IS the ecdsa address by defn)

src/ECDSASignatureChecker.sol
```diff
@@ -0,0 +1,170 @@
...
+            // The check below validates that operatorIds are sorted (and therefore free of duplicates)
+            if (i != 0) {
+                require(
+                    uint256(signerIds[i]) >  uint256(signerIds[i - 1]),
+                    "ECDSASignatureChecker.checkSignatures: signer keys not sorted"
+                );
+            }
+
+            // check the operator's signature
+            EIP1271SignatureUtils.checkSignature_EIP1271(address(uint160(uint256(signerIds[i]))), msgHash, signatures[i]);
```
samlaf, line 125: related to my previous comments, if we store the operatorId as operatorAddr instead (literally type `address` instead of `bytes32`), we don't need to do this conversion here (`address(uint160(uint256))`). Afaik there's no other use for storing the operatorId as a bytes32 instead of an address?

samlaf: Changed it in my branch and everything checked out.

src/ECDSARegistryCoordinator.sol
```diff
@@ -0,0 +1,867 @@
...
+import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
+import {IServiceManager} from "./interfaces/IServiceManager.sol";
+
+import {BitmapUtils} from "./libraries/BitmapUtils.sol";
+
+/**
+ * @title A `RegistryCoordinator` that has three registries:
+ *      1) a `StakeRegistry` that keeps track of operators' stakes
+ *      2) a `BLSApkRegistry` that keeps track of operators' BLS public keys and aggregate BLS public keys for each quorum
+ *      3) an `IndexRegistry` that keeps track of an ordered list of operators for each quorum
```
samlaf, line 23: Was trying to add an OperatorStateRetriever for ecdsa inc-sq and realized its no longer possible (without the indexRegistry) to get a list of operators.

samlaf (commented):
Sorry for bombarding you @ChaoticWalrus with my every passing thoughts :P

Seems like the main problem I'm facing right now is resulting from the combination of:
1. no "epoch" history (only the current state of oeprators is stored in contracts)
2. no "epoch" timeline - aka permissionless entry/exit, at any moment
So any avs would be totally broken if they start aggregating signatures offchain (think eigenDA) and then some operator joins/leaves.

This is related to the new avs-sync proposal I've done for churning operators. Basically there seems to be an epoch spectrum:
"fluid" (reactive) epochs -------------------------> hard epochs (after P&S)
current eigenDA                                                     bft chain with epoch-based consensus

If we are moving to a hard-epoch based system after P&S, feels like eigenDA will also need to follow suit and adapt (but maybe not?). I was thinking new operators joining could just be added to the next epoch's list, and we change the list pointer (similar to doing a green-blue deployment) at the epoch boundary... but maybe this doesn't work for smart contracts as it would require another tx for all the operators?

Would like some help thinking through this in more detail whenever you have time.

ChaoticWalrus:
> Sorry for bombarding you @ChaoticWalrus with my every passing thoughts :P
> 
> Seems like the main problem I'm facing right now is resulting from the combination of:
> 
>     1. no "epoch" history (only the current state of oeprators is stored in contracts)
> 
>     2. no "epoch" timeline - aka permissionless entry/exit, at any moment
>        So any avs would be totally broken if they start aggregating signatures offchain (think eigenDA) and then some operator joins/leaves.
> 
> 
> This is related to the new avs-sync proposal I've done for churning operators. Basically there seems to be an epoch spectrum: "fluid" (reactive) epochs -------------------------> hard epochs (after P&S) current eigenDA bft chain with epoch-based consensus
> 
> If we are moving to a hard-epoch based system after P&S, feels like eigenDA will also need to follow suit and adapt (but maybe not?). I was thinking new operators joining could just be added to the next epoch's list, and we change the list pointer (similar to doing a green-blue deployment) at the epoch boundary... but maybe this doesn't work for smart contracts as it would require another tx for all the operators?
> 
> Would like some help thinking through this in more detail whenever you have time.

You have a ton of ideas here; agreed we should chat about them.
I'll try to give the bulk of your ideas thought here -- I've read what you've been writing but haven't poured more effort into this.
To respond to one thing in particular:

>        So any avs would be totally broken if they start aggregating signatures offchain (think eigenDA) and then some operator joins/leaves.

Why do you think this? If you submit a bunch of ECDSA signatures at once and an operator isn't registered, just add 0 to the total "stake/weight" here (or otherwise skip some calculation steps). If an operator newly registered and you don't have their signature...there's no issue I can see at all.
These race condition kind of problems are actually way worse with the BLS scheme we're using and were a motivator for a ton of the historical storage of our contracts, which is part of why I aggressively ripped historical storage out in this PR.

src/ECDSARegistryCoordinator.sol
ChaoticWalrus: ahahahah 😅 
OK guess I went a little too hard on removing things.
good to point out the need for this (or equivalent)
but if we are doing a "historyless" implementation then we could likely have a much-more-minimal version of the IndexRegistry
or you could just say "this is designed for users to have to index the operators off-chain" and the equivalent OperatorStateRetriever would require you to input specific operator addresses

samlaf:
> Why do you think this? If you submit a bunch of ECDSA signatures at once and an operator isn't registered, just add 0 to the total "stake/weight" here (or otherwise skip some calculation steps). If an operator newly registered and you don't have their signature...there's no issue I can see at all.

wdym by "just add 0"? where would this be? in the contract? or in the aggregator offchain code?
The way inc-sq works is that a task request is submitted to the contract, and the task response is accepted only if the threshold is met wrt the stake weight of operators **at the time the task was submitted**. Given that we don't have history of these updates, if a new operator joins with a bunch of stake, then that totally changes the distrbution, so a signature that would have been accepted might get rejected.

ChaoticWalrus:
> wdym by "just add 0"? where would this be? in the contract? or in the aggregator offchain code?
The way inc-sq works is that a task request is submitted to the contract, and the task response is accepted only if the threshold is met wrt the stake weight of operators at the time the task was submitted. Given that we don't have history of these updates, if a new operator joins with a bunch of stake, then that totally changes the distrbution, so a signature that would have been accepted might get rejected.

ahhhh I missed this latter point but you're totally right. But TBH I think these issues might be acceptable -- just gather 60% of the signatures instead of 50% or w/e and the probability that your tx reverts is tiny

samlaf:
I implemented an ecdsa aggregation service in services/ecdsa/aggregation/agg.go and started wondering whether there's a way to do ecdsa verification without having a centralized aggregator, aka every operator just signs and sends their signature to the contract. Seems like this would require more gas because each time a new signature comes the contract needs to calculate and store the current stake, until it meets threshold. With update histories like I used in that PR (which uses the [ecdsa-with-history-poc](https://github.com/Layr-Labs/eigenlayer-middleware/tree/ecdsa-with-history-poc/src/experimental) branch) the contract would also need to be passed for each validator its update history indices to fetch its correct stake, and we'd have to make sure all operators are using the same referenceBlockNumber.

Just curious in general what are people's thoughts around this, since aggregating in the smart contract seems like an easy way to get rid of the centralized aggregator role (wihout having to deal with a p2p network and leader aggregating - what happens if a task is not finished aggregating and then leader changes, etc). I think gnosis safes also work with a centralized aggregator though?

ChaoticWalrus:
> I implemented an ecdsa aggregation service in services/ecdsa/aggregation/agg.go and started wondering whether there's a way to do ecdsa verification without having a centralized aggregator, aka every operator just signs and sends their signature to the contract. Seems like this would require more gas because each time a new signature comes the contract needs to calculate and store the current stake, until it meets threshold. With update histories like I used in that PR (which uses the [ecdsa-with-history-poc](https://github.com/Layr-Labs/eigenlayer-middleware/tree/ecdsa-with-history-poc/src/experimental) branch) the contract would also need to be passed for each validator its update history indices to fetch its correct stake, and we'd have to make sure all operators are using the same referenceBlockNumber.
> 
> Just curious in general what are people's thoughts around this, since aggregating in the smart contract seems like an easy way to get rid of the centralized aggregator role (wihout having to deal with a p2p network and leader aggregating - what happens if a task is not finished aggregating and then leader changes, etc). I think gnosis safes also work with a centralized aggregator though?

Technically Gnosis Safes support both, but their UI uses a centralized aggregator (at least by default).
I think "aggregatorless" solutions would be great to develop though! I can see some of how it interacts with other architecture choices too.
Overall I would expect this type of thing to cost more gas pretty much regardless, so in terms of prioritization I'd be curious to understand the demand here.

src/ECDSARegistryCoordinator.sol
```diff
@@ -0,0 +1,867 @@
...
+        return results;
+    }
+
+    function _getOrCreateOperatorId(
+        address operator,
+        ECDSAPubkeyRegistrationParams memory params
+    ) internal returns (bytes32 operatorId) {
+        operatorId = _operatorInfo[operator].operatorId;
+        if (operatorId == 0) {
+            operatorId = _registerECDSAPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
```
samlaf, line 558: ~~I think this is a bug, and should be `pubkeyRegistrationMessageHash(params.signingAddress)` since this is verifying a signature against the params address, not the operator address. Can you certify whether I'm correct @ChaoticWalrus , need this for my branch as well.~~ EDIT: I was totally wrong

Btw to differentiate the 2 addresses in ecdsa inc-sq, I've been using `operatorEcdsaAddress`, and `avsEcdsaAddress`. Perhaps we should stick to `operatorId` instead of `avsEcdsaAddress`, but it feels easier at least for the offchain code to prevent bugs by calling them the `operator` key and the `avs` key (meaning the key that is used to sign msgs offchain).

samlaf: hmmm nvm I think I'm wrong actually. Somehow the checkSignature is failing for me saying I have the wrong signer... let me double check.

ChaoticWalrus: idk on naming. IMO depends on your intended usage.
I think `operatorId` is more general, but `avsEcdsaAddress` is more specific. Probably needs `ECDSA` in there somewhere, assuming you need different handlers for these things

ChaoticWalrus: I don't know if I had thought the details of this through, but this method only gets run when called by the operator, and if the operator chooses a bad `signingAddress` that _could_ just be placed on them.
The signature here I guess is ultimately just confirming that the `signingAddress` "agrees" to the arrangement.

ChaoticWalrus: One more thought on naming: I'll note as well that since we're using EIP1271 signature checking here, the `signingAddress` could be a contract that does whatever custom signature checking the operator wants, so it doesn't _strictly_ need to be ECDSA

samlaf: Super helpful comments! Will resolve nonetheless since my first point was just wrong.

ChaoticWalrus:
I think this was a useful exercise but am not seeing taking this particular branch/PR further, at least for months+.
Seems like https://github.com/Layr-Labs/eigenlayer-middleware/pull/194 accomplishes the goals of this PR, but more fully (albeit with other changes).
So I will close this PR for the moment; the git history will continue to exist, should anyone ever want to pick this particular direction up again.

PR #124, opened 2023-12-28
feat: example sig checker loop collapse

wadealexc:


wadealexc:
Closing as impractical :)

Issue #130, opened 2024-01-01
Test: integration/fuzz tests


wadealexc:
## Description

Core-style integration testing, deploying a full AVS and mocking as little as possible, then running through various scenarios and fuzzing some of the more complicated pathways.

### Action Items
- [x] Build deployment/test setup
- [x] Identify some scenarios or specific paths in the code that should be tested
- [x] Build integration test framework a la core contracts
- [x] Write tests
- [x] PR

PR #132, opened 2024-01-02
test: add integration test framework and basic integration test

wadealexc:
Scenario document here: https://hackmd.io/@-HV50kYcRqOjl_7du8m1AA/BySD_8Eda

Check `tests/integration/README.md` for documentation


wadealexc:
Related: https://github.com/Layr-Labs/eigenlayer-middleware/issues/130

samlaf:
Just saw this. Have you seen our previous requirement/design doc for this? https://docs.google.com/document/d/1wrPNkvOyAtxGXB34crLgFKTUU8O56ceOgvjyZg7oGPs/edit#heading=h.gaj8gxkqakmv

wadealexc:
> Just saw this. Have you seen our previous requirement/design doc for this? https://docs.google.com/document/d/1wrPNkvOyAtxGXB34crLgFKTUU8O56ceOgvjyZg7oGPs/edit#heading=h.gaj8gxkqakmv

No, but that doc is significantly more involved than I have time for.

The goal of this PR is to mock out user flows through the middleware contracts based on a handful of scenarios, and fuzz test those flows across a variety of random quorum configs

samlaf:
> The goal of this PR is to mock out user flows through the middleware contracts based on a handful of scenarios, and fuzz test those flows across a variety of random quorum configs

ah so its more of an onchain integration test then? Only checking integration of the different contracts together?
Whereas I was talking about integration test with some offchain code.

wadealexc:
> > The goal of this PR is to mock out user flows through the middleware contracts based on a handful of scenarios, and fuzz test those flows across a variety of random quorum configs
> 
> ah so its more of an onchain integration test then? Only checking integration of the different contracts together? Whereas I was talking about integration test with some offchain code.

That is correct!

test/integration/IntegrationBase.t.sol
```diff
@@ -0,0 +1,920 @@
...
+
+        uint96[] memory curTotalStakes = _getTotalStakes(churnedQuorums);
+        uint96[] memory prevTotalStakes = _getPrevTotalStakes(churnedQuorums);
+
+        // For each quorum, check that the incoming operator's individual stake was increased by addedWeights
+        // and that the total stake is plus addedWeights and minus removedWeights
+        for (uint i = 0; i < churnedQuorums.length; i++) {
+            assertEq(curIncomingOpStakes[i], prevIncomingOpStakes[i] + addedWeights[i], err);
+            assertEq(curTotalStakes[i], prevTotalStakes[i] + addedWeights[i] - removedWeights[i], err);
+        }
```
ChaoticWalrus, line 361: I guess you could add a check that the churned operator's stakes are now zero in the quorums?
perhaps overkill though, as this is coming quite close already 😄

test/integration/IntegrationChecks.t.sol
```diff
@@ -0,0 +1,418 @@
...
+        assert_HasAtLeastMinimumStake(incomingOperator, combinedQuorums,
+            "operator should have at least the minimum stake in each quorum");
+        assert_Snap_Added_OperatorWeight(incomingOperator, standardQuorums,
+            "failed to add operator weight to operator and total stake in standardQuorums");
+        assert_Snap_Churned_OperatorWeight(incomingOperator, churnedOperators, churnedQuorums,
+            "failed to add operator weight and remove churned weight from each quorum");
+
+        // IndexRegistry
+        assert_Snap_Added_OperatorCount(standardQuorums,
+            "total operator count should have increased for standardQuorums");
```
ChaoticWalrus, line 123: noting that this _could_ break in some circumstances, since standard quorums will have operators deregister to make room.
doesn't look to be breaking now, so probably a moot point though.

test/integration/IntegrationBase.t.sol
wadealexc: To check the snapshot version of that, this method would need to do one additional timewarp lookup per churned operator, which really slows down the tests.

I do a non-timewarp version of that in `check_Churned_State`, instead, which is where this method is called

test/integration/IntegrationChecks.t.sol
wadealexc: This is true, but they were deregistered before the last snapshot was taken.

The snap only checks the prior state right before the user called registerOperatorWithChurn

test/integration/IntegrationChecks.t.sol
```diff
@@ -0,0 +1,418 @@
...
+
+        // Core
+        assert_Snap_Added_OperatorShares(operator, strategies, tokenBalances,
+            "operator should have additional stake");
+    }
+
+    /// @dev Checks that an operator's stake was successfully increased
+    /// NOTE: This method assumes (and checks) that the operator already
+    ///       met the minimum stake before stake was added.
+    function check_DepositUpdate_State(
```
ChaoticWalrus, line 207: perhaps more name differentiation between this method and `check_Deposit_State` could be good?
I have yet to look through the usage but it seems like `check_Deposit_State` is meant to be used after a "deposit to core" and then `check_DepositUpdate_State` is meant to be used after a follow-up "stake update" is triggered on the AVS?
Perhaps `check_DepositCore_State` and `check_DepositStakeUpdate_State` or something might be clearer?

ChaoticWalrus: or maybe just another comment could help. idk. things are clean, there's just a lot to keep straight!

edit: I think the equivalent of the documentation on the `check_Withdraw_XYZ` methods is really what I'm looking for here, tbh

wadealexc: There are a handful of unclear names, so I appreciate feedback like this. Will do!

test/integration/IntegrationChecks.t.sol
ChaoticWalrus: I was about to note that my understanding of the snapshot state was wrong when I made this comment, glad to see you corrected me already.
resolving now as I agree that this is good as-is!

ChaoticWalrus (approved):
This PR is 🔥 🔥 🔥 
Huge kudos as this is a massive amount of work, and a very cool framework.

Gets me excited for other things we can do!
I'd love to see us add some coverage for config changes (like changing quorum params or adding additional quorums) as a follow-up to this, in particular.

test/integration/IntegrationBase.t.sol
ChaoticWalrus: makes sense! yeah I would just leave as-is then.

wadealexc:
>  I'd love to see us add some coverage for config changes (like changing quorum params or adding additional quorums) as a follow-up to this, in particular.

Me too, but i gotta shift to audit prep! maybe next week

PR #135, opened 2024-01-04
Update BLSApkRegistry.sol: Optimized gas

0xScratch:
This PR aims for two changes:

1. **Enhanced Modifier using internal functions** - Can be used when any modifier is used often in a contract..These are pretty good for references - https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#use-internal-view-functions-in-modifiers-to-save-bytecode, https://www.rareskills.io/post/gas-optimization#viewer-527f6

2. **Enhanced For loops** - After solidity version 0.8, solidity uses an integer overflow/underflow checker which costs some gas..but whenever we are sure about our function, `unchecked` blocks can be used to save some gas..This will help - https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement

Thanks! @wadealexc @ChaoticWalrus

### DIFF
--- a/src/BLSApkRegistry.sol
+++ b/src/BLSApkRegistry.sol
@@ -12,10 +12,7 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
 
     /// @notice when applied to a function, only allows the RegistryCoordinator to call it
     modifier onlyRegistryCoordinator() {
-        require(
-            msg.sender == address(registryCoordinator),
-            "BLSApkRegistry.onlyRegistryCoordinator: caller is not the registry coordinator"
-        );
+        _checkRegistryCoordinator();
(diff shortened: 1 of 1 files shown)


wadealexc:
Thanks for the PR -- 

1. `uint i = 0;` and `uint i;` are the same gas cost
2. we removed the `unchecked` portions of the code as it's easier to accidentally skip index incrementing when you use them. The gas savings aren't large enough to make them worth it
3. I'd like to keep the modifier you changed as-is; IIRC you're correct that this saves bytecode, but I prefer the explicit in-line syntax.

0xScratch:
okay 👍

PR #138, opened 2024-01-09
feat: operator set root

0x0aa0:
Adds a running hash of stake updates for use in storage proof


src/StakeRegistry.sol
```diff
@@ -273,6 +273,14 @@ contract StakeRegistry is StakeRegistryStorage {
         }
     }
 
+    /**
+     * @notice clears the quorum root for the given quorum
+     * @dev This function is called by the registry coordinator on stake updates to consolidate the size of the storage proof
+     */
+    function clearQuorumRoot(uint8 quorumNumber) external onlyRegistryCoordinator quorumExists(quorumNumber) {
```
gpsanant, line 280: emit an event

src/StakeRegistry.sol
```diff
@@ -495,6 +508,12 @@ contract StakeRegistry is StakeRegistryStorage {
         return _totalStakeHistory[quorumNumber].length != 0;
     }
 
+    /// @notice Updates the quorum root for the given quorum 
+    /// @dev quorumRoot+1 = h(quorumRoot, {operator stake update})
+    function _updateQuorumRoot(bytes32 operatorId, uint8 quorumNumber, uint32 blockNumber, uint96 stake) internal {
```
gpsanant: just hash the stake update itself imo, not the individual components

src/StakeRegistryStorage.sol
```diff
@@ -43,6 +43,9 @@ abstract contract StakeRegistryStorage is IStakeRegistry {
      */
     mapping(uint8 => StrategyParams[]) public strategyParams;
 
+    /// @notice mapping from quorum number to the root used for storage proofs of operator set 
+    mapping(uint8 => bytes32) public quorumRoot;
```
gpsanant: quorumOperatorSetRoot

src/StakeRegistry.sol
0x0aa0: it will always hash with `nextUpdateBlockNumber: 0` in that case

gpsanant: not when they're deregistering?

src/StakeRegistry.sol
```diff
@@ -347,6 +359,8 @@ contract StakeRegistry is StakeRegistryStorage {
                     nextUpdateBlockNumber: 0,
                     stake: newStake
                 }));
+
+                _updateQuorumRoot(operatorId, quorumNumber, uint32(block.number), newStake);
```
wadealexc: Should be called below, before the function returns. Currently, this misses a branch where the operator's stake is updated twice in one block.

src/StakeRegistry.sol
wadealexc: > not when they're deregistering?

also when they're deregistering tho. `nextUpdateBlockNumber` is only set to a nonzero value when we push a new one - so if the hash is over the most recent update, the next update block number will be zero.

src/RegistryCoordinator.sol
```diff
@@ -323,6 +323,7 @@ contract RegistryCoordinator is
                 "RegistryCoordinator.updateOperatorsForQuorum: number of updated operators does not match quorum total"
             );
             address prevOperatorAddress = address(0);
+            stakeRegistry.clearQuorumRoot(quorumNumber);
```
wadealexc, line 315: Are there any cases where we don't want this to be called?

This call here basically means anyone can clear the current quorum root on demand. Is that ok?

src/StakeRegistryStorage.sol
```diff
@@ -43,6 +43,9 @@ abstract contract StakeRegistryStorage is IStakeRegistry {
      */
     mapping(uint8 => StrategyParams[]) public strategyParams;
 
+    /// @notice mapping from quorum number to the root used for storage proofs of operator set 
+    mapping(uint8 => bytes32) public quorumOperatorSetRoot;
```
8sunyuan, line 47: decrement storage gap for this new mapping slot?

src/RegistryCoordinator.sol
0x0aa0: Yes this ok left permissionless as long as the root is always repopulated with stake updates. This refreshing serves to reduce the size of the corresponding storage proof

src/StakeRegistry.sol
```diff
@@ -359,11 +357,19 @@ contract StakeRegistry is StakeRegistryStorage {
                     nextUpdateBlockNumber: 0,
                     stake: newStake
                 }));
-
-                _updateQuorumRoot(operatorId, quorumNumber, uint32(block.number), newStake);
             }
         }
 
+        _updateQuorumRoot(
```
gpsanant, line 363: you also need to update where the operator deregisters?

wadealexc: This is the method that gets called when they deregister

wadealexc: That's why arch was saying this only gets hashed with nextUpdateBlockNumber == 0

src/StakeRegistry.sol
gpsanant: oh ok then. no need

src/StakeRegistry.sol
gpsanant: responded above. sg.

0x0aa0: _recordOperatorStakeUpdate() is also called on dereregistration

src/StakeRegistry.sol
```diff
@@ -350,6 +360,16 @@ contract StakeRegistry is StakeRegistryStorage {
...
+            quorumNumber, 
+            operatorId, 
+            StakeUpdate({
+                updateBlockNumber: uint32(block.number),
+                nextUpdateBlockNumber: 0,
```
gpsanant, line 368: prob don't need this but its ok

gpsanant (approved):
lgtm

wadealexc:
changed the PR title so the linter is happier when merged

teddyknox:
LGTM! Will verify that storage proofs match `<StakeRegistry contract address>.quorumOperatorSetRoot[quorumNumber]`, and then will verify that the value matches `keccak256(abi.encode(quorumOperatorSetRoot[quorumNumber], operatorId, operatorStakeUpdate))`.

teddyknox (approved):

wadealexc (changes requested):
Per conversation in Slack, we're going to wait to merge this until it's well tested and documented.

PR #139, opened 2024-01-11
chore: repo cleanup

### LINKED ISSUE #126: IndexRegistry does not update `currentOperatorIndex` when an operator is deregistered, in certain cases
Came across this while working on documentation: 

https://github.com/Layr-Labs/eigenlayer-middleware/blob/ebe657e1167e59dc7959dfe11c8194b72bf9a347/src/IndexRegistry.sol#L99-L106

If `operatorId == lastOperatorId`, we never assign `currentOperatorIndex[operatorId]`.

To fix this, we can change `_popLastOperator` or add an `else` condition

### LINKED ISSUE #127: Cleanup: unify typing of state variables in StakeRegistryStorage
https://github.com/Layr-Labs/eigenlayer-middleware/blob/ebe657e1167e59dc7959dfe11c8194b72bf9a347/src/StakeRegistryStorage.sol#L30-L35

This is the only place in the middleware contracts where we use `[256]` to represent quorum information. I'd like to replace this with a mapping for consistency. I don't think it actually impacts a single thing, but if it this ends up breaking things it's fine to keep as-is.

### LINKED ISSUE #131: Split RegistryCoordinator storage into RegistryCoordinatorStorage contract
This is the only contract we have that doesn't use the same storage pattern as the rest of the contracts. We should split out storage and include a `_gap` variable for consistency.

### LINKED ISSUE #133: Add docs/comments for RegistryCoordinator.initialize
https://eigenlabs.slack.com/archives/C05PX94L3H9/p1703619404949479

### LINKED ISSUE #134: Clarify comments for BitmapUtils.plus and minus methods
Current commenting is unclear, and should reflect that impl is closer to "union" / "negate + intersect" operations

### LINKED ISSUE #136: BLSSignatureChecker.checkSignatures: validate that input `quorumNumbers` is non-empty
Ref: https://github.com/Layr-Labs/eigenlayer-middleware/blob/m2-mainnet/src/BLSSignatureChecker.sol#L86

wadealexc:
* closes #126
* closes #127
* closes #131
* closes #133
* closes #134
* closes #136

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -46,6 +46,8 @@ contract RegistryCoordinator is
         _;
     }
 
+    /// @dev Checks that `quorumNumber` corresponds to a quorum that has been created
+    /// via `initialize` or `createQuorum`
     modifier quorumExists(uint8 quorumNumber) {
         require(
             quorumNumber < quorumCount, 
@@ -66,6 +68,17 @@ contract RegistryCoordinator is
         _disableInitializers();
     }
 
+    /**
+     * @param _initialOwner will hold the owner role
+     * @param _churnApprover will hold the churnApprover role, which authorizes registering with churn
+     * @param _ejector will hold the ejector role, which can force-eject operators from quorums
+     * @param _pauserRegistry a registry of addresses that can pause the contract
+     * @param _initialPausedStatus pause status after calling initialize
+     * Config for initial quorums (see `createQuorum`):
+     * @param _operatorSetParams max operator count and operator churn parameters
+     * @param _minimumStakes minimum stake weight to allow an operator to register
+     * @param _strategyParams which Strategies/multipliers a quorum considers when calculating stake weight
+     */
     function initialize(
         address _initialOwner,
         address _churnApprover,
@@ -103,13 +116,14 @@ contract RegistryCoordinator is
     *******************************************************************************/
 
     /**
-     * @notice Registers msg.sender as an operator for one or more quorums. If any quorum reaches its maximum
-     * operator capacity, this method will fail.
+     * @notice Registers msg.sender as an operator for one or more quorums. If any quorum excees its maximum
+     * operator capacity after the operator is registered, this method will fail.
      * @param quorumNumbers is an ordered byte array containing the quorum numbers being registered for
-     * @param socket is the socket of the operator
+     * @param socket is the socket of the operator (typically an IP address)
      * @param params contains the G1 & G2 public keys of the operator, and a signature proving their ownership
-     * @dev the `params` input param is ignored if the caller has previously registered a public key
      * @param operatorSignature is the signature of the operator used by the AVS to register the operator in the delegation manager
+     * @dev `params` is ignored if the caller has previously registered a public key
+     * @dev `operatorSignature` is ignored if the operator's status is already REGISTERED
      */
     function registerOperator(
         bytes calldata quorumNumbers,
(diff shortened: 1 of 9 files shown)


src/RegistryCoordinator.sol
```diff
@@ -152,13 +116,14 @@ contract RegistryCoordinator is
     *******************************************************************************/
 
     /**
-     * @notice Registers msg.sender as an operator for one or more quorums. If any quorum reaches its maximum
-     * operator capacity, this method will fail.
+     * @notice Registers msg.sender as an operator for one or more quorums. If any quorum excees its maximum
+     * operator capacity after the operator is registered, this method will fail.
```
ChaoticWalrus: typo: `excees` => `exceeds`

src/RegistryCoordinator.sol
```diff
@@ -275,11 +248,10 @@ contract RegistryCoordinator is
...
 
     /**
-     * @notice Updates the stakes of one or more operators in the StakeRegistry, for each quorum
-     * the operator is registered for.
-     * 
-     * If any operator no longer meets the minimum stake required to remain in the quorum,
-     * they are deregistered.
+     * @notice Updates the StakeRegistry's view of one or more operators' stakes. If any operator
+     * is found to be below the minimum stake for the quorum, they are deregistered.
+     * @dev stakes are queryed from the Eigenlayer core DelegationManager contract
```
ChaoticWalrus: quark-sized nit but I believe it should be `queried` instead of `queryed`

ChaoticWalrus: same comment on `updateOperatorsForQuorum` 😅

src/RegistryCoordinator.sol
```diff
@@ -295,51 +267,72 @@ contract RegistryCoordinator is
...
-     * @dev This method is used to update the stakes of all operators in a quorum at once, rather than individually. Performs
-     * sanitization checks on the input array lengths, quorumNumbers existing, and that quorumNumbers are ordered. Function must
-     * also not be paused by the PAUSED_UPDATE_OPERATOR flag.
+     * @notice For each quorum in `quorumNumbers`, updates the StakeRegistry's view of ALL its registered operators' stakes.
+     * Each quorum's `quorumUpdateBlockNumber` is also updated, which tracks the most recent block number when ALL registered
+     * operators were updated.
+     * @dev stakes are queryed from the Eigenlayer core DelegationManager contract
+     * @param operatorsPerQuorum for each quorum in `quorumNumbers`, this has a corresponding list of operators to update.
+     * @dev Each list of operator addresses MUST be sorted in ascending order
+     * @dev Each list of operator addresses MUST represent the entire list of registered operators for the corresponding quorum
```
ChaoticWalrus, line 276: might not hurt to acknowledge the potential race condition here? (i.e. if an operator registers/deregisters, it can cause this to revert)

src/RegistryCoordinator.sol
```diff
@@ -503,6 +511,24 @@ contract RegistryCoordinator is
         return operatorId;
     }
 
+    /**
+     * @notice Validates that an incoming operator is eligible to replace an existing
+     * operator based on the stake of both
+     * @dev In order to churn, the incoming operator needs to have more stake than the
+     * existing operator by a proportion given by `kickBIPsOfOperatorStake`
+     * @dev In order to be churned, the existing operator needs to have less stake than
+     * the total quorum stake by a propertion given by `kickBIPsOfTotalStake`
```
ChaoticWalrus: typo in `propertion`
I think this wording may also be a little confusing?
I would say instead:
```
     * @dev In order to be churned out, the existing operator needs to have a proportion of the total
     * quorum stake less than `kickBIPsOfTotalStake`
```
but perhaps that's just my personal preference

src/RegistryCoordinator.sol
```diff
@@ -555,30 +581,31 @@ contract RegistryCoordinator is
...
+        // Deregister operator with each of the registry contracts
         blsApkRegistry.deregisterOperator(operator, quorumNumbers);
         stakeRegistry.deregisterOperator(operatorId, quorumNumbers);
         indexRegistry.deregisterOperator(operatorId, quorumNumbers);
     }
 
     /**
-     * @notice update operator stake for specified quorumsToUpdate, and deregister if necessary
-     * does nothing if operator is not registered for any quorums.
+     * @notice Updates the StakeRegistry's view of the operator's stake one or more quorums.
```
ChaoticWalrus: I feel like there's a word missing. perhaps "_in_ one or more quorums"?

ChaoticWalrus (approved):
LGTM! I only have a few really minor, typo-level comments

PR #143, opened 2024-01-11
refactor: add operator address to getOperatorState

8sunyuan:
Adding operator address to `Operator` struct in the stateRetriever to make it easier to call `RegistryCoordinator.updateOperators` offchain for avs-sync.


samlaf (approved):
LGTM. thanks for this!

PR #146, opened 2024-01-15
feat: automated ejector

0x0aa0:
Here we introduce the EjectionManager contract, the purpose of which is to enforce a ratelimit on stake for ejections called automatically by an offchain component. The contract prioritizes successful execution over the race conditions of changes in total stake and regular operator deregistration.


src/Ejector.sol
```diff
@@ -0,0 +1,149 @@
...
+import {IEjector} from "./interfaces/IEjector.sol";
+import {IRegistryCoordinator} from "./interfaces/IRegistryCoordinator.sol";
+import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
+import {BitmapUtils} from "./libraries/BitmapUtils.sol";
+
+/**
+ * @title Used for automated ejection of operators from the RegistryCoordinator
+ * @author Layr Labs, Inc.
+ */
+contract Ejector is IEjector, Ownable{
```
gpsanant: should be upgradeable

src/interfaces/IEjector.sol
```diff
@@ -0,0 +1,61 @@
...
+pragma solidity ^0.8.9;
+
+import {IRegistryCoordinator} from "./IRegistryCoordinator.sol";
+import {IStakeRegistry} from "./IStakeRegistry.sol";
+
+/**
+ * @title Interface for a contract that ejects operators from an AVSs RegistryCoordinator
+ * @author Layr Labs, Inc.
+ */
+interface IEjector {
```
gpsanant: EjectionManager might be better to differentiate from the "ejector" role

src/interfaces/IEjector.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    /// @notice A quorum's ratelimit parameters
+    struct QuorumEjectionParams {
+        uint32 timeDelta; // Time delta to track ejection over
+        uint16 ejectableStakePercent; // Max stake to be ejectable per time delta
```
gpsanant: `rateLimitWindow`?

src/Ejector.sol
```diff
@@ -0,0 +1,149 @@
...
+                require(quorumEjectionParams[quorumNumber].timeDelta > 0, "Ejector: Quorum ejection params not set");
+
+                uint256 operatorStake = stakeRegistry.getCurrentStake(_operatorIds[i], quorumNumber);
+
+                if(msg.sender == ejector){
+                    _cleanOldEjections(quorumNumber);
+                    require(canEject(operatorStake, quorumNumber), "Ejector: Stake exceeds quorum ejection ratelimit");
+                }
+
+                stakeEjectedForQuorum[quorumNumber].push(StakeEjection({
```
gpsanant: we can for sure make this 1 sstore?

src/Ejector.sol
```diff
@@ -0,0 +1,149 @@
...
+            quorumEjectionParams[i] = _quorumEjectionParams[i];
+        }
+    }
+
+    /**
+     * @notice Ejects operators from the AVSs registryCoordinator
+     * @param _operatorIds The addresses of the operators to eject
+     * @param _quorumBitmaps The quorum bitmaps for each respective operator
+     */
+    function ejectOperators(bytes32[] memory _operatorIds, uint256[] memory _quorumBitmaps) external {
```
gpsanant: i prefer something like
`uint256 quorumBitmap, bytes32[quorumNumbers][numOperators] operatorIds`
then we sum up the stake of operators for each quorum and boot them out if low enough and time delta has passed.

seems much simpler and just need to store the previous ejection time

src/EjectionManager.sol
```diff
@@ -0,0 +1,170 @@
...
+            _setQuorumEjectionParams(i, _quorumEjectionParams[i]);
+        }
+    }
+
+    /**
+     * @notice Ejects operators from the AVSs registryCoordinator under a ratelimit
+     * @dev This function will eject as many operators as possible without reverting
+     * @param _operatorIds The ids of the operators to eject for each quorum
+     */
+    function ejectOperators(bytes32[][] memory _operatorIds) external {
```
gpsanant, line 65: i feel like `ejectOperators(uint8 quorumNumber, bytes32[] calldata _operatorIds)` is more in line with our style

src/EjectionManager.sol
```diff
@@ -0,0 +1,170 @@
...
+                ){
+                    stakeEjectedForQuorum[quorumNumber].push(StakeEjection({
+                        timestamp: block.timestamp,
+                        stakeEjected: stakeForEjection
+                    }));
+                    broke = true;
+                    break;
+                }
+                
+                try registryCoordinator.ejectOperator(
```
gpsanant, line 93: why would we not revert if this failed? why would this fail?

0x0aa0: Would fail if the operator leaves the set before ejection is called

src/EjectionManager.sol
0x0aa0: https://github.com/Layr-Labs/eigenlayer-middleware/pull/146#discussion_r1456445550

src/EjectionManager.sol
gpsanant: ah gpgp

src/EjectionManager.sol
```diff
@@ -0,0 +1,163 @@
...
+        _setEjector(_ejector);
+
+        for(uint8 i = 0; i < _quorumEjectionParams.length; i++) {
+            _setQuorumEjectionParams(i, _quorumEjectionParams[i]);
+        }
+    }
+
+    /**
+     * @notice Ejects operators from the AVSs RegistryCoordinator under a ratelimit
+     * @param _operatorIds The ids of the operators to eject for each quorum
```
jianoaix: What's assumed in this parameter by this function?
From the impl below it looks: 1) the caller provides all operators eligible for ejection, but delegate to this function for rate limiting; 2) the operators in each quorum are ordered by priority to eject (in case not all can be ejected, it'll just eject a leading prefix of the array)
Might be good to document the assumptions so the caller knows how to pass the params.

src/EjectionManager.sol
```diff
@@ -0,0 +1,168 @@
...
+            uint256 stakeForEjection; 
+
+            bool broke;
+            for(uint8 j = 0; j < _operatorIds[i].length; ++j) {
+                uint256 operatorStake = stakeRegistry.getCurrentStake(_operatorIds[i][j], quorumNumber);
+
+                //if caller is ejector enforce ratelimit
+                if(
+                    msg.sender == ejector && 
+                    quorumEjectionParams[quorumNumber].rateLimitWindow > 0 &&
```
ChaoticWalrus, line 81: seems like this means that no limit will be enforced for new quorums until a limit is explicitly set on this contract for the quorum?
possibly a bit of a "footgun"?
please correct me if I'm reading this wrong though.

src/EjectionManager.sol
```diff
@@ -0,0 +1,168 @@
...
+        }
+        while(stakeEjectedForQuorum[_quorumNumber][i].timestamp > cutoffTime) {
+            totalEjected += stakeEjectedForQuorum[_quorumNumber][i].stakeEjected;
+            if(i == 0){
+                break;
+            } else {
+                --i;
+            }
+        }
+        return totalEjectable - totalEjected;
```
ChaoticWalrus, line 170: is there a chance of (reverting on) underflow here, e.g. if `ejectableStakePercent` is edited?
seems fine but might be a rather opaque failure?

ChaoticWalrus (approved):
LGTM overall. Noted a couple more minor edge cases / quirks.
I reviewed the tests briefly but not in detail.

src/EjectionManager.sol
0x0aa0: good call

src/EjectionManager.sol
0x0aa0: That is correct. My thinking here is that when adding a new quorum, while it is finding its footing we will not want to run automated ejection but reserve the ability to eject without ratelimit in that period. Once the set is more stable the quorum params can be set and offchain can be turned on to run regularly.

src/EjectionManager.sol
gpsanant: you caught me red handed. now i realize why i wanted that is we want to eject from multiple quorums at the same time

gpsanant: we could do that outside of the contract too, but ok

src/EjectionManager.sol
```diff
@@ -0,0 +1,168 @@
...
+     * @param _quorumNumber The quorum number to view ejectable stake for
+     */
+    function amountEjectableForQuorum(uint8 _quorumNumber) public view returns (uint256) {
+        uint256 cutoffTime = block.timestamp - quorumEjectionParams[_quorumNumber].rateLimitWindow;
+        uint256 totalEjectable = quorumEjectionParams[_quorumNumber].ejectableStakePercent * stakeRegistry.getCurrentTotalStake(_quorumNumber) / BIPS_DENOMINATOR;
+        uint256 totalEjected;
+        uint256 i;
+        if (stakeEjectedForQuorum[_quorumNumber].length == 0) {
+            return totalEjectable;
+        } else {
```
gpsanant: nit: take out of else

src/EjectionManager.sol
```diff
@@ -0,0 +1,168 @@
...
+            for(uint8 j = 0; j < _operatorIds[i].length; ++j) {
+                uint256 operatorStake = stakeRegistry.getCurrentStake(_operatorIds[i][j], quorumNumber);
+
+                //if caller is ejector enforce ratelimit
+                if(
+                    msg.sender == ejector && 
+                    quorumEjectionParams[quorumNumber].rateLimitWindow > 0 &&
+                    stakeForEjection + operatorStake > amountEjectable
+                ){
+                    stakeEjectedForQuorum[quorumNumber].push(StakeEjection({
```
gpsanant, line 84: why are we pushing and not continuing here?

0x0aa0: Because we do not want to revert if the stake of the ops in the input is greater than the ratelimit. Here we checkpoint the stake ejected inside the ratelimit and then end execution

gpsanant: oh my bad, i completely misread this

PR #148, opened 2024-01-18
chore: remove unused/outdated files

8sunyuan:
- Removing `VoteWeigherBaseUnit.t.sol`, these tests are no longer needed as the Voteweighing contract was consolidated into the StakeRegistry and the unit tests from this file have already been migrated in `StakeRegistryUnit.t.sol`
- `src/interfaces/IDelayedService.sol` is also not being used anywhere so removing this interface


ChaoticWalrus (approved):
LGTM

PR #155, opened 2024-01-26
chore: storage gaps and nits

stevennevins:
This PR addresses the following discussed on slack

- Makes the `ServiceManagerBase` abstract
- Replaces `ServiceManagerBase` in tests with a `ServiceManagerMock`
- Adds an internal ServiceManagerBase init function that has an `onlyInitializing` modifier so it can be called within a function that has an `initializer` modifier
- Removes the `initialize` function since it likely won't be used by inheriting contracts
- Adds storage gaps to the `ServiceManagerBase` and `BLSSignatureChecker` which are likely to be inherited by upgradeable contracts. and any changes to their storage layouts could have a downstream impact without these gaps.


ChaoticWalrus (approved):
Looks C L E A N
nice changes, thanks for suggesting these and getting them in.

wadealexc (approved):
LGTM

PR #162, opened 2024-01-30
fixes(m2-mainnet): combined pr for all m2-mainnet fixs

stevennevins:


wadealexc (approved):
LGTM! Would be good to get another signoff before merging.

(and we may want to merge after core)

ChaoticWalrus:
> we may want to merge after core

I'm unsure on the best order-of-operations here, myself

I do think it might be best to the following sequencing:

1. merge core repo
2. update submodule commit for the core repo, on the `m2-mainnet-fixes` branch of this repo
3. merge this PR

but I don't see a reason we _can't_ merge this PR right now, so I think it's fine, at least.

ChaoticWalrus (approved):
Looks good.
Seems likely this will need follow-up changes to deal with the analogous merging process in the core repo(?)

PR #163, opened 2024-01-30
test: fix flaky tests by removing bogosort

### LINKED ISSUE #161: Fix flaky test by sorting operatorIds
`BLSMockAVSDeployer` does horrible things to generate a sorted list of random operator ids:

https://github.com/layr-labs/eigenlayer-middleware/blob/cac457271cc43feb8aefe886ef359a4f6621cb9c/test/utils/BLSMockAVSDeployer.sol#L57-L67

This results in tests failing very rarely due to "out of gas". We should sort operatorIds here rather than hashing elements.

wadealexc:
- closes #161


ChaoticWalrus (approved):
Bogosort is best sort, but I can abide this change

PR #165, opened 2024-02-01
feat: add experimental, epoch-based ECDSA Registry

ChaoticWalrus:
Disclaimer: this could very well be broken in its current form

This POC-level product is permissioned and minimally-featured

However, it manages to combine down all Registry functionality (sans what's been deleted, ofc)
into a single, reasonably-sized contract
which is also hopefully easier to understand

To provide more details on how this works:
Operators can register or deregister "at any time", which sets a flag and passes a call onto the ServiceManager
Regardless of these actions, the operator set (you can think of this as the "active" operators) for an epoch **does not change** -- it is established once for the epoch, and after that the set and the weights of operators are fixed for the remainder of the epoch.
For establishing the operator set:
currently the call is permissioned, which is perhaps one of the main shortcomings here
at the start of each epoch, the contract owner provides a list of operator addresses, and a minimum weight requirement
https://github.com/Layr-Labs/eigenlayer-middleware/pull/165/files#diff-b6eec5ea8b39de39a8dc5657b3cc4dfacc5a99b95a71dc182d8174a866a0cd3aR190-R210
any operators in that list which are actively registered + meet the requirement go into the set of operators for the epoch.


ChaoticWalrus:
Some updates on recent progress:

1. The [SimpleLinearWeightQuorum](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/SimpleLinearWeightQuorum.sol) takes inspiration from steven’s simplifications to the Quorum config [here](https://github.com/Layr-Labs/eigencert/blob/feat/updates/contracts/src/ECDSAStakeRegistry.sol) — there’s now just a [single function](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/SimpleLinearWeightQuorum.sol#L60-L62) for doing all config changes (add strategy, adjust weight, etc.)

2. The [EcdsaEpochRegistry_Modular](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/EcdsaEpochRegistry_Modular.sol) just combines what I was doing before with this — operators can register whenever they want, owner has to evaluate the operator set at the start of every epoch (trusted party inputs the set of registered operators to potentially include in the new active set for the epoch)

3. The [EcdsaEpochRegistry_Permissionless](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/EcdsaEpochRegistry_Permissionless.sol) adds an exponential retargeting mechanism (a la 1559) — the `minimumWeightRequirement` is adjusted every epoch to (using a `retargetingFactorWei`  and the real number of registered operators vs. the target).
Any operator which meets the `minimumWeightRequirement` can register, up to `maxRegisteredOperators`
At the start of each epoch, anyone (now not just the contract owner) can perform the operation to find the validator set for the epoch (and kick out any operators that no longer meet the  `minimumWeightRequirement`. once this is done the retargeting is triggered once again.
I am 1000% sure that the algorithm can be improved, but I think this is a cool direction to go.

4. [EcdsaEpochBase](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/EcdsaEpochBase.sol) takes the above and just adds a simple `checkSignatures` [function](https://github.com/Layr-Labs/eigenlayer-middleware/blob/c2444acac13024b041737e66f195c762c386d339/src/experimental/EcdsaEpochBase.sol#L47-L56). this contract is `9.165kB` in size, which means it could itself be the “ServiceManager” for an AVS, with ~15kB remaining for AVS developers to add whatever other functionality they want.

Some directions for improvement / additions:

- better retargeting algorithm
- add in some “operator performance / reputation” or other kind of metric to the weighting
- add an optional “ejector” and/or blacklist/banning mechanism
- fuller implementation of task confirmation (e.g. storing signatories hash)
- code cleanup/de-duplication; better modularization in general
- support for multiple quorums (perhaps with different config-setters)

src/experimental/EcdsaEpochBase.sol
```diff
@@ -0,0 +1,82 @@
...
+            // The check below validates that operatorIds are sorted (and therefore free of duplicates)
+            if (i != 0) {
+                require(
+                    operators[i] >  operators[i - 1],
+                    "ECDSASignatureChecker.checkSignatures: signer keys not sorted"
+                );
+            }
+
+            // check the operator's signature
+            // TODO: any modifications to msgHash?
```
stevennevins: wdym here?

ChaoticWalrus: it wasn't immediately clear to me if we should e.g. require operators to sign something like `keccack256(msgHash, operator)` instead of just having them sign the `msgHash` directly

stevennevins: Ahh I see

How you have it should be sufficient

ChaoticWalrus:
Other things to add:

- indexing support (e.g. make it easier to determine the operator set for each historical epoch)
- support operators signing with a key other than their address itself (currently this branch simplifies and just has operators signing with the key of their address)
- possibly return the new operator set for the epoch at the end of calculating it?
- better events in general
- multiquorum support -- in particular with access control where a different entity can control the config of each quorum

ChaoticWalrus:
~~So... I have rebased this branch twice now, and the diff here still shows a bunch of independent changes.
I will likely have to regen this branch; seems like maybe things got messed up when the default branch was changed or something?~~

edit: I appear to have salvaged things by clicking a button 😄

ChaoticWalrus:
Added a `StandaloneQuorum` contract, which could be shared between different AVSs -- there is no notion of "quorum membership" but AVSs could use this to outsource quorum definition + simply query weights when they want them; by outsourcing these calculations, if AVSs share a StandaloneQuorum, then the costs of weight calculations can be amortized.
Copy-pasting the top-level explanatory comment I wrote for the contract:

/**
 * @notice This contract is similar to SimpleLinearWeightQuorum but with a staleness measure, `maxWeightStalenessBlocks`.
 * Provides a few main methods: `getOperatorWeight`, `updateWeightOfOperatorIfNecessary`, and `forceOperatorWeightUpdate`
 * -- see these functions for more details on usage & behavior.
 * Adds one owner-only function for modifying the value of `maxWeightStalenessBlocks`.
 */

ChaoticWalrus:
closing this PR in favor of https://github.com/Layr-Labs/eigenlayer-middleware/pull/209

PR #169, opened 2024-02-02
test: fix flaky fuzz test

stevennevins:
- Fixes a flaky fuzz test that occurs intermittently or at high fuzz runs, which causes a different revert reason than the one we `vm.expect`


stevennevins:
@8sunyuan just bumping this

PR #170, opened 2024-02-02
test/refactor: gas scenarios for updateOperators

8sunyuan:
Created a draft PR to get some updated avs-sync gas costs.
The added integration tests gives the following results:

10 Operators; 25 strats; Gas cost: 849185
For 200 operators: 849185*20 = 16,983,700

10 Operators; 20 strats; Gas cost: 699895
For 200 operators: 699895*20 = 13,997,900

10 Operators; 15 strats; Gas cost: 550605
For 200 Operators: 550605*20 = 11,012,100

**After Using getOperatorShares() in _weightOfOperator()** 
Turns out using one external call to DM to get an operators weight greatly reduces gas cost by ~23%

10 Operators; 25 strats; Gas cost: 657245
For 200 operators: 657245*20 = 13,144,900

10 Operators; 20 strats; Gas cost: 553025
For 200 operators: 553025*20 = 11,060,500

10 Operators; 15 strats; Gas cost: 448815
For 200 Operators: 448815*20 = 8,976,300


src/StakeRegistry.sol
```diff
@@ -418,6 +418,7 @@ contract StakeRegistry is StakeRegistryStorage {
                 "StakeRegistry._addStrategyParams: cannot add strategy with zero weight"
             );
             strategyParams[quorumNumber].push(_strategyParams[i]);
+            strategiesPerQuorum[quorumNumber].push(_strategyParams[i].strategy);
```
wadealexc, line 424: Well, I wish we weren't duplicating storage here but... this is a much simpler change than splitting `strategyParams` into 2 arrays.

And I assume we have backend stuff relying on `strategyParams` and at this point we'd like to not change that.

wadealexc (commented):
One thing I notice about this -- this might cause our goerli code to break, because it doesn't have `strategiesPerQuorum` populated.

test/integration/IntegrationConfig.t.sol
```diff
@@ -488,6 +498,12 @@ contract IntegrationConfig is IntegrationDeployer, G2Operations, Constants {
             strategyCount = 2;
         } else if (strategyFlag == MANY) {
             strategyCount = _randUint({ min: 3, max: allStrats.length - 1 });
+        } else if (strategyFlag == FIFTEEN) {
+            strategyCount = 15;
+        } else if (strategyFlag == TWENTY) {
+            strategyCount = 20;
+        } else if (strategyFlag == TWENTYFIVE) {
+            strategyCount = 25;
```
wadealexc, line 506: Were these added for the gas tests?

test/mocks/DelegationMock.sol
```diff
@@ -89,6 +89,11 @@ contract DelegationMock is IDelegationManager {
         return 0;
     }
 
+    function getOperatorShares(
+        address operator,
+        IStrategy[] memory strategies
+    ) external view returns (uint256[] memory) {}
```
wadealexc: I'm gonna guess that this needs to return an array with equal length to `strategies`, and that's why a lot of tests are panicing with arr out of bounds

wadealexc: actually, tried this out - i think this just needs to be replaced with the actual method rather than an empty method. tests are passing now

src/StakeRegistryStorage.sol
```diff
@@ -42,6 +42,8 @@ abstract contract StakeRegistryStorage is IStakeRegistry {
      * corresponding multipliers for that specific quorum
      */
     mapping(uint8 => StrategyParams[]) public strategyParams;
+    mapping(uint8 => IStrategy[]) public strategiesPerQuorum;
```
ChaoticWalrus, line 45: gotta reduce the storage gap size by 1 if we add this new mapping

src/StakeRegistry.sol
ChaoticWalrus: oof yes agreed this is ugly but I can see the reasoning.
will need to add other parallel storage updates to make these 2 lists consistent, as right now this PR appears to just be duplicating adding entries (but not removing them).

one option could be to split `strategyParams` into 2 arrays, but support a view function which combines an entry from each array to maintain interface compatibility -- but I assume this still would break indexing, so...worst of both worlds perhaps.

src/StakeRegistryStorage.sol
wadealexc: we really need to add some automated checking for that after we do mainnet

src/StakeRegistry.sol
wadealexc, line 424: Secondary note - need to ensure strategies are removed from this list in `removeStrategies`!

wadealexc: oh nvm i see jeffrey is saying that above

wadealexc: > will need to add other parallel storage updates to make these 2 lists consistent

pulling this out -- `removeStrategies` needs to delete entries from this list

8sunyuan: Yeah as this was a last minute optimization onchain, I figured we shouldn't change the current storage layout here. Perhaps the next testnet deployment this could be updated?

wadealexc: More likely a future upgrade entirely - this is more a question of whether we want to go tell backend/infra to update our indexer, IIUC.

It's really not a huge deal in the grand scheme of things. Doesn't make the average user's gas costs worse, just ours (and only by a marginal amount on a method we won't be calling much).

test/integration/IntegrationConfig.t.sol
8sunyuan: Yup, tests seem to be ~12M gas for 200 operators and 25 strategies for updateOperators

8sunyuan:
Also, updated eigenlayer-contracts commit hash to be `751155931cfb6b5ebcb77ef3cb12dff041329cbc`
can be viewed here https://github.com/Layr-Labs/eigenlayer-contracts/commit/751155931cfb6b5ebcb77ef3cb12dff041329cbc

src/StakeRegistry.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/170/commits/3688a02f4a7841f2d3befbdd3821cc3b4d22c602#diff-ff6dec6ce26deefc6d5cb801b1d654b03cba3bd352386f0a828990d979a97332R241
Fixed here for removing strategies

test/mocks/DelegationMock.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/170/commits/3688a02f4a7841f2d3befbdd3821cc3b4d22c602#diff-7501823751a3937eed2beae0fb540d70cf5b4abae705e6c6d7a49ccc4e7942dfR95
Good callout, added

src/StakeRegistryStorage.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/170/commits/3688a02f4a7841f2d3befbdd3821cc3b4d22c602#diff-3aca207bfe24bd32c66e23f704da75c45051c4fad2b745aa2efaf937e9dc6e2dR58

8sunyuan:
> One thing I notice about this -- this might cause our goerli code to break, because it doesn't have `strategiesPerQuorum` populated.

Per slack, DA contracts will be redeployed and EL will be upgraded so all good!

wadealexc (approved):
LGTM

PR #176, opened 2024-02-09
Update docs with goerliv2 deployment

0x0aa0:
Adds addresses for latest testnet deployment


wadealexc:
Closing since we're now pointing at the public docs site :(

PR #181, opened 2024-02-12
fix: prevent use of current block as reference block number

### LINKED ISSUE #171: BLSSignatureChecker::checkSignatures - Prevent current block from being used as `referenceBlockNumber`
From Middleware audit, items P1, M1, and L2:

Although `BLSSignatureChecker::checkSignatures` checks that the reference block number is not in the future, the current block (i.e., the next to be produced) is accepted as reference:

```solidity
function checkSignatures(
	bytes32 msgHash,
	bytes calldata quorumNumbers,
	uint32 referenceBlockNumber,
	NonSignerStakesAndSignature memory params
){
	...
	require(referenceBlockNumber <= uint32(block.number), "BLSSignatureChecker.checkSignatures: invalid reference block");
```

As a consequence, users of eigenlayer-middleware could be tempted to submit confirmations for the current block, in an attempt to provide them as early as possible. This practice, however, is risky, as it creates race conditions between the signature check and other updates that might be performed in the same block.

If an update happens in the same block before checkSignatures, it might violate some of the data included in the check, such as the APK or the stake history indexes, causing the signature check to fail. In issues M1 and L2 we discuss in detail two scenarios in which such a race condition could be caused maliciously to achieve a DoS attack.

Similarly, an update might happen in the same block but after checkSignatures. In this case, the signature check will succeed, but there will be an inconsistency between the data being checked and the information stored for that block, which could have undesirable consequences for the protocol. For instance, the signatoryRecord hashed and stored by EigenDAServiceManager::confirmBatch could be referring to a block with data inconsistent with the globally-maintained data about that block. To reconstruct records, one would need to replay past transactions.

As a consequence we recommend strengthening the above require to ensure that only blocks strictly in the past can be used as reference. Alternatively, AVS implementations should be careful to ensure they can tolerate such races and apparent-inconsistency in records.

wadealexc:
- Closes #171 
- Addresses Dedaub M1, L2, P1

### DIFF
--- a/src/BLSSignatureChecker.sol
+++ b/src/BLSSignatureChecker.sol
@@ -109,7 +109,7 @@ contract BLSSignatureChecker is IBLSSignatureChecker {
             "BLSSignatureChecker.checkSignatures: input nonsigner length mismatch"
         );
 
-        require(referenceBlockNumber <= uint32(block.number), "BLSSignatureChecker.checkSignatures: invalid reference block");
+        require(referenceBlockNumber < uint32(block.number), "BLSSignatureChecker.checkSignatures: invalid reference block");
 
         // This method needs to calculate the aggregate pubkey for all signing operators across
         // all signing quorums. To do that, we can query the aggregate pubkey for each quorum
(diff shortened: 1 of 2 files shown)


ChaoticWalrus (approved):
LGTM

PR #182, opened 2024-02-12
fix: include missing fields in TYPEHASH defintion

ChaoticWalrus:
the `salt` and `expiry` fields were missing from the `OPERATOR_CHURN_APPROVAL_TYPEHASH` def

### DIFF
--- a/src/RegistryCoordinatorStorage.sol
+++ b/src/RegistryCoordinatorStorage.sol
@@ -15,7 +15,7 @@ abstract contract RegistryCoordinatorStorage is IRegistryCoordinator {
 
     /// @notice The EIP-712 typehash for the `DelegationApproval` struct used by the contract
     bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
-        keccak256("OperatorChurnApproval(address registeringOperator,bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams,bytes32 salt,uint256 expiry)OperatorKickParam(address operator,bytes32[] operatorIdsToSwap)");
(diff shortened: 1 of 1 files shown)


src/RegistryCoordinatorStorage.sol
```diff
@@ -15,7 +15,7 @@ abstract contract RegistryCoordinatorStorage is IRegistryCoordinator {
 
     /// @notice The EIP-712 typehash for the `DelegationApproval` struct used by the contract
     bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
-        keccak256("OperatorChurnApproval(address registeringOperator,bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams)OperatorKickParam(address operator,bytes32[] operatorIdsToSwap)");
+        keccak256("OperatorChurnApproval(address registeringOperator,bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams,bytes32 salt,uint256 expiry)OperatorKickParam(address operator,bytes32[] operatorIdsToSwap)");
```
wadealexc: Looks like the definition of OperatorKickParam is incorrect

wadealexc (changes requested):
Just need to correct `OperatorKickParam` and we're good!

src/RegistryCoordinatorStorage.sol
ChaoticWalrus: really, really good catch! thank you! it should be fixed as of https://github.com/Layr-Labs/eigenlayer-middleware/pull/182/commits/c543bc9f190e148b9da6d52a92a05c3f9381653b

wadealexc (approved):
LGTM!

PR #185, opened 2024-02-13
feat: basic ecdsa stake registry

stevennevins:


src/unaudited/ECDSAStakeRegistry.sol
samlaf, line 1: Can you add a README in this unaudited directory outlining the relationship between these 4 different files?

Seems like the hierarchy is
EqualWeights -> Permissioned -> StakeRegistry -> StakeRegistryStorage
?

Unobvious why that is the case when first coming to this so would help to have a bit of prose to form a quick mental model.

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+
+import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import {CheckpointsUpgradeable} from "@openzeppelin-upgrades/contracts/utils/CheckpointsUpgradeable.sol";
+import {PausableUpgradeable} from "@openzeppelin-upgrades/contracts/security/PausableUpgradeable.sol";
+import {SignatureCheckerUpgradeable} from "@openzeppelin-upgrades/contracts/utils/cryptography/SignatureCheckerUpgradeable.sol";
+import {IERC1271Upgradeable} from "@openzeppelin-upgrades/contracts/interfaces/IERC1271Upgradeable.sol";
+
+/// @title ECDSA Stake Registry
+/// @notice THIS CONTRACT IS NOT AUDITED
+/// @notice Manages operator registration and quorum updates for a staking system using ECDSA signatures.
```
samlaf: what is a "staking system"? Should we say an AVS instead?

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+/// @dev Extends OpenZeppelin's upgradeable Ownable and Pausable patterns to ensure upgradability and pause functionality.
+contract ECDSAStakeRegistry is
+    IERC1271Upgradeable,
+    OwnableUpgradeable,
+    PausableUpgradeable,
+    ECDSAStakeRegistryStorage
+{
+    using SignatureCheckerUpgradeable for address;
+    using CheckpointsUpgradeable for CheckpointsUpgradeable.History;
```
samlaf, line 27: Could you add a comment above this to explain to newcomers what exactly these checkpoints are and how we are using it? Was confusing to me the first time I saw this. Unless this is explained elsewhere.

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+    function deregisterOperator() external {
+        _deregisterOperator(msg.sender);
+    }
+
+    /**
+     * @notice Updates the StakeRegistry's view of one or more operators' stakes.
+     * @dev Queries stakes from the Eigenlayer core DelegationManager contract
+     * @param _operators A list of operator addresses to update
+     */
+    function updateOperators(address[] memory _operators) external {
```
samlaf, line 62: This naming will break avs-sync afaiu, because avs-sync calls `updateOperatorsForQuorums` when it wants to update the entire operator set (`updateOperators` is meant to update a small subset, say an operator who just wants to update his own position).

Maybe we should still call it `updateOperatorsForQuorums` here, even though there's a single quorum? And explain why in a @dev note? Maybe we can also find a better naming convention for the bls registryCoordinator as well.. not sure..

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    /// @notice Updates the weight an operator must have to join the operator set
+    /// @dev Access controlled to the contract owner
+    /// @param _newMinimumWeight The new weight an operator must have to join the operator set
+    function updateMinimumWeight(uint256 _newMinimumWeight) external onlyOwner {
+        minimumWeight = _newMinimumWeight;
+    }
+
+    /**
+     * @notice Adjusts the threshold weight for valid signatures.
```
samlaf: is this for valid signatures? Or is this the threshold OF weights of signers at which an "aggregate" signature is accepted?

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+            _signatureData,
+            (address[], bytes[], uint32)
+        );
+        _checkSignatures(_dataHash, signers, signatures, referenceBlock);
+        return IERC1271Upgradeable.isValidSignature.selector;
+    }
+
+    /// @notice Retrieves the current stake quorum details.
+    /// @return Quorum - The current quorum of strategies and weights
+    function stakeQuorum() external view returns (Quorum memory) {
```
samlaf: why is this called stakeQuorum as opposed to just quorum? Might make users wonder if this is different from the other quorums we are using elsewhere?

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    /**
+     * @notice Adjusts the threshold weight for valid signatures.
+     * @param _thresholdWeightBps The updated threshold weight.
+     */
+    function updateStakeThreshold(uint256 _thresholdWeightBps) external onlyOwner {
+        _updateStakeThreshold(_thresholdWeightBps);
+    }
+
+    function isValidSignature(
```
samlaf, line 90: nit: add docs

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+        uint32 _blockNumber
+    ) external view returns (uint256) {
+        return _totalWeightHistory.getAtBlock(_blockNumber);
+    }
+
+    function operatorRegistered(address _operator) external view returns (bool) {
+        return _operatorRegistered[_operator];
+    }
+
+    /// @notice Calculates the current weight of an operator based on their delegated stake in the strategies considered in the quourm
```
samlaf: quourm -> quorum

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    /// @notice Deregisters an existing operator
+    function deregisterOperator() external {
+        _deregisterOperator(msg.sender);
+    }
+
+    /**
+     * @notice Updates the StakeRegistry's view of one or more operators' stakes.
+     * @dev Queries stakes from the Eigenlayer core DelegationManager contract
```
samlaf, line 59: Can we also add comments as to how this is supposed to be used. Like explain the intended use. Reading this contract quickly, unclear to me if this is supposed to be called once per epoch (is this even an epoch based contract?). Or can updateOperators be called whenever and it just creates a new checkpoint for each validator?

samlaf (commented):
first quick pass of comments.

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: This is just attaching the library to the struct. The variables below are named relative to what they represent

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Hmm, so we only update in AVS-sync in per quorum basis? This function is similar to the one that exists on the registry coordinator already that I thought we supported

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: it’s not really an aggregate signature, but its the threshold stake that all the signers combined must meet

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Will need to update the internal variable. It conflicted so I just named it stakeQuorum

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: No epoch, you can just call it and add an updated view of the operator

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: It’s not that equal weights are the intended version to be used. It’s just a variant of the registry that teams have indicated interest in.

Any of the 3 could be a starting point to use or starting point to customize

src/unaudited/ECDSAStakeRegistry.sol
samlaf: ah gotcha my bad.

src/unaudited/ECDSAStakeRegistry.sol
samlaf: `updateOperators` - meant to be called by a single operator (or small oeprator group) to update their entire set of quorums
`updateOperatorsForQuorums` - meant to be called by AVS teams to update their entire set of operators once per quorum. It's meant to be called individually per quorum because a single quorum could be close to max gas per block.

src/unaudited/ECDSAStakeRegistry.sol
samlaf: Right, just saying the comment is inaccurate and leads to confusion. It sounds like the threshold weight applies to a single signature. Something like "Adjusts the threshold weight required to accept a group of signatures" maybe?

src/unaudited/ECDSAStakeRegistry.sol
samlaf: Ok how about a comment such as
```
Calling this will update the stake weight of all operators passed as parameters, 
adding a new entry in their history of stake checkpoints, 
effectively entering a new "epoch" for this set of operators.
```

src/unaudited/ECDSAStakeRegistry.sol
samlaf: sgtm. Just document this in a README to make it easy for people to reason about these.

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: makes sense, i'll add that function as well

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: I will add clarity, but the epoch usage here needs to be more accurate. Maybe for the entire quorum update function that would work

PR #187, opened 2024-02-16
Draft: ServiceManager Router

ypatil12:


ChaoticWalrus (commented):
Seems sensible, looks good, really appreciate that you also wrote tests for this.
Perhaps this should be in the core repo? But I suppose the test files it reuses make more sense in this repo so... super cool with it staying here.

PR #188, opened 2024-02-16
chore: update core submodule and fix broken docs links

wadealexc:


ChaoticWalrus (approved):
Everything looks correct to me. Thanks for this!

PR #190, opened 2024-02-20
update eigenlayer-contracts to head of dev to get changed to deploy_from_scratch script

samlaf:


samlaf:
@stevennevins can you check into the failing test. Doesn't seem to be because of my changes.

stevennevins (approved):

PR #194, opened 2024-02-26
feat: ecdsa basic registry

stevennevins:
This PR introduces a simpler registry implementation as an alternative to the existing StakeRegistry.sol.  This is based on the teams asking for more examples and a general requirement to use ECDSA as their signature scheme.

Also, standardizing to use an `isValidSignature` function is proposed to query if a set of signatures from operators are valid at a given block height which isn't contained in the other implementation but seems like a useful standardization


src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+    function updateQuorumConfig(Quorum memory _quorum) external onlyOwner {
+        _updateQuorumConfig(_quorum);
+    }
+
+    /// @notice Updates the weight an operator must have to join the operator set
+    /// @dev Access controlled to the contract owner
+    /// @param _newMinimumWeight The new weight an operator must have to join the operator set
+    function updateMinimumWeight(uint256 _newMinimumWeight) external onlyOwner {
+        minimumWeight = _newMinimumWeight;
+    }
```
ChaoticWalrus: probably want an event for this as well

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+    /// @param _operator The address of the operator.
+    /// @return uint256 - The current weight of the operator; returns 0 if below the threshold.
+    function getOperatorWeight(address _operator) public view returns (uint256) {
+        StrategyParams[] memory strategies = _quorum.strategies;
+        uint256 weight;
+        uint256 sharesAmount;
+        for (uint256 i; i < strategies.length; i++) {
+            sharesAmount = DELEGATION_MANAGER.operatorShares(_operator, strategies[i].strategy);
+            weight += sharesAmount * strategies[i].multiplier;
+        }
```
ChaoticWalrus: there's now an arrayified version of this function which should be more efficient to call https://github.com/Layr-Labs/eigenlayer-contracts/blob/dev/src/contracts/core/DelegationManager.sol#L864-L873

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+    /// @param _newQuorum The new quorum configuration to set
+    function _updateQuorumConfig(Quorum memory _newQuorum) internal {
+        if (!_isValidQuorum(_newQuorum)) revert InvalidQuorum();
+        Quorum memory oldQuorum = _quorum;
+        delete _quorum;
+        for (uint256 i; i < _newQuorum.strategies.length; i++) {
+            _quorum.strategies.push(_newQuorum.strategies[i]);
+        }
+        emit QuorumUpdated(oldQuorum, _newQuorum);
+    }
```
ChaoticWalrus: seems like this is only updating the `strategies` and not the `weights` ?

I took some inspiration from your struct definition + wrote a similar function here https://github.com/Layr-Labs/eigenlayer-middleware/blob/970349466df340dc87c79fb3029f37c6b464a646/src/experimental/SimpleLinearWeightQuorum.sol#L59-L86

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+    /**
+     * @notice Updates the StakeRegistry's view of one or more operators' stakes adding a new entry in their history of stake checkpoints, 
+     * @dev Queries stakes from the Eigenlayer core DelegationManager contract
+     * @param _operators A list of operator addresses to update
+     */
+    function updateOperators(address[] memory _operators) external {
+        for (uint256 i; i < _operators.length; i++) {
+            _updateOperatorWeight(_operators[i]);
+        }
+    }
```
ChaoticWalrus: seems like there needs to be a check for the operators being registered. perhaps safest to check in the internal function?

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Need to look into the origin of the naming here, but the `strategies` array is an array of `strategyParams`.  The multiplier in the `strategyParmas` is effectively the weight

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: yeah nice catch

stevennevins: added a regression case for this

stevennevins: I realized this was already handled in the `_updateOperatorWeight` function where operators that aren't registered either because they were deregistered immediately before the `_updateOperatorWeight` call or because they weren't registered in the first place can only push a weight of 0 to their history.

https://github.com/Layr-Labs/eigenlayer-middleware/blob/948538b93971c8384500e35a5ad07dfe0d2ccd88/src/unaudited/ECDSAStakeRegistry.sol#L230

The clarity of this code path could be improved, though

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/842b403713e8222ade39eb2b03c7c68b5890c3de

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+        address lastStrategy;
+        address currentStrategy;
+        uint256 totalMultiplier;
+        for (uint256 i; i < strategies.length; i++) {
+            currentStrategy = address(strategies[i].strategy);
+            if (lastStrategy >= currentStrategy) revert NotSorted();
+            lastStrategy = currentStrategy;
+            totalMultiplier += strategies[i].multiplier;
+        }
+        if (totalMultiplier != BPS) return false;
```
ChaoticWalrus: hmm, interesting idea. seems odd though that adding strategies would require reducing the multipliers for existing strategies? 
I think this could have negative side effects with asynchronous weight updates to operators, where e.g. the same operator gets a lower weight after a change even though they have the same amount of stake delegated to them

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+    ) internal pure {
+        if (_signersLength != _signaturesLength) revert LengthMismatch();
+        if (_signersLength == 0) revert InvalidLength();
+    }
+
+    /// @notice Ensures that signers are sorted in ascending order by address.
+    /// @param _lastSigner The address of the last signer.
+    /// @param _currentSigner The address of the current signer.
+    function _validateSortedSigners(address _lastSigner, address _currentSigner) internal pure {
+        if (_lastSigner >= _currentSigner) revert NotSorted();
```
ChaoticWalrus: you want strict inequality here to prevent duplicates, not non-strict inequality

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,385 @@
...
+            return _thresholdWeightBpsHistory.getAtBlock(_referenceBlock);
+        }
+    }
+
+    /// @notice Validates that the cumulative stake of signed messages meets or exceeds the required threshold.
+    /// @param _signedWeight The cumulative weight of the signers that have signed the message.
+    /// @param _referenceBlock The block number to verify the stake threshold for
+    function _validateThresholdStake(uint256 _signedWeight, uint32 _referenceBlock) internal view {
+        uint256 totalWeight = _getTotalWeight(_referenceBlock);
+        if (_getThresholdStake(_referenceBlock) > (_signedWeight * BPS) / totalWeight)
```
ChaoticWalrus: probably want non-strict inequality here?
also given the usage of `_thresholdWeightBpsHistory` you might want to sanitize those inputs more, so the threshold can't exceed 100%

src/unaudited/examples/ECDSAStakeRegistryPermissioned.sol
```diff
@@ -0,0 +1,48 @@
...
+    /// @param _operator The address of the operator to be allowlisted
+    function permitOperator(address _operator) external onlyOwner {
+        allowlistedOperators[_operator] = true;
+    }
+
+    /// @notice Revokes an operator's permission and deregisters them
+    /// @param _operator The address of the operator to remove from the allowlist and deregistered.
+    function revokeOperator(address _operator) external onlyOwner {
+        delete allowlistedOperators[_operator];
+        _deregisterOperator(_operator);
```
ChaoticWalrus: seems like this line would end up reverting if the `_operator` is not actively registered.
not a major issue but would recommend addressing this

ChaoticWalrus (commented):
LGTM overall. seems like some reasonable examples to illustrate a few ideas.
I noted several things but am not seeing any big, architectural issues; it all seems logically organized in general, and very similar to existing code but making more use of OZ in particular.

src/unaudited/ECDSAStakeRegistry.sol
ChaoticWalrus: ahhhhh my bad! OK will resolve thread.

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/33627dca38d71ad814717e1d2dfd888eacdc2873

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Yeah, that's a good point in this construction. I was trying to cache some operations in the update stage and was thinking you would just couple the quorum strategy weight updates with the updateOperators, but it does lead to a discrepancy in your weight if a strategy you were heavily weighted in has its multiplier decreased and you haven't updated the operator to reflect the change.

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Strict inequality would be if I inverted the logic and used a `require`. I had a NotSorted test and added a Duplicate test just to be thorough https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/55da82a13b0f97fb4068f6b63f96111b33712e98

src/unaudited/examples/ECDSAStakeRegistryPermissioned.sol
stevennevins: yeah nice callout, updated https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/a72f7e910be201ce010c23387d817ef9053b1253

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: I think the inequality choice is correct.

if `signedWeight` == `thresholdWeight`, then don't revert
if `signedWeight` > `thresholdWeight`, then don't revert
if `signedWeight` < `thresholdWeight`, then revert

For sanitization, do you mean make sure that `_signedWeight` is less than or equal to `totalWeight`?

stevennevins: It seems like this should only be a concern in the "out of sync" scenario discussed for updating the multipliers on strategies within a quorum and not subsequently updating all the operators' weights.  I will add the check against totalWeight here, but it seems like there also needs to be an update to all the operators when the quorum changes to make sure there is never more `signedWeight` than `totalWeight`

stevennevins: Sanitizing change https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/8843c90cf332996ca25461a748a4e9974a242c25

Will follow up with the additional change discussed to update all the operator weights when the quorum changes

stevennevins: Follow up to make sure weights are always in sync with the global configuration for the quorum
https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/f723e4f61d4985e7c706564f023b32f51af30f0f

src/unaudited/ECDSAStakeRegistry.sol
ChaoticWalrus: ah, mb, it's just from custom error logic being flipped

src/unaudited/ECDSAStakeRegistry.sol
ChaoticWalrus: I again read this wrong, totally my bad.
for input sanitization I believe I was thinking that _updateStakeThreshold needs to make sure its input is always less than or equal to `BPS` -- my apologies for the lack of clarity there -- I think this is the right place to sanitize:
https://github.com/Layr-Labs/eigenlayer-middleware/blob/f723e4f61d4985e7c706564f023b32f51af30f0f/src/unaudited/ECDSAStakeRegistry.sol#L196-L199

ChaoticWalrus:
Thanks for addressing so many of my comments! There's a couple threads still open; let's sync up on where this is headed

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: this invariant should address this [97592a6](https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/97592a677513db4c0791741e47014ae58433d30c)

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: makes sense, yeah this should have a check to make sure the new `_thresholdWeightBps` is less than `BPS`

src/unaudited/ECDSAStakeRegistry.sol
ChaoticWalrus: seems reasonable. totally avoids what I'm saying as long as the whole operator set is updated at once

src/unaudited/examples/ECDSAStakeRegistryPermissioned.sol
```diff
@@ -0,0 +1,50 @@
...
+    /// @param _operator The address of the operator to be allowlisted
+    function permitOperator(address _operator) external onlyOwner {
+        allowlistedOperators[_operator] = true;
+    }
+
+    /// @notice Revokes an operator's permission and deregisters them
+    /// @param _operator The address of the operator to remove from the allowlist and deregistered.
+    function revokeOperator(address _operator) external onlyOwner {
+        delete allowlistedOperators[_operator];
+        if (_operatorRegistered[_operator]){
```
afkbyte: think this check is redundant, `_deregisterOperator` seems to already perfom it

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,414 @@
...
+    /// @param _operatorSignature Contains the operator's signature, salt, and expiry
+    function registerOperatorWithSignature(
+        address _operator,
+        ISignatureUtils.SignatureWithSaltAndExpiry memory _operatorSignature
+    ) external {
+        _registerOperatorWithSig(_operator, _operatorSignature);
+    }
+
+    /// @notice Deregisters an existing operator
+    function deregisterOperator() external {
```
afkbyte, line 56: I'm curious if there's any cases where we would not want an operator to be able to deregister themselves from an AVS

stevennevins: yeah, I could see there being restrictions on when they could deregister themselves i.e, if they're still under some bonding commitment. but the scenarios would probably be highly specific to the AVS

src/unaudited/examples/ECDSAStakeRegistryPermissioned.sol
stevennevins: We need this check; otherwise, we will revert to the check you're talking about if the operator isn't registered.  

We could probably have two functions, `revokeOperator` and `revokeAndEjectOperator`, so we could remove checking the value twice, but I think gas here isn't too important since it's an admin function

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: improved the threshold weight usage in this commit https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/e360e91a54ed668fd5311b03edc68436b47884b6

NimaVaziri:
@stevennevins seems like @ChaoticWalrus is good with this / the remaining comments are minor - any blockers to get this over the line?

stevennevins:
> @stevennevins seems like @ChaoticWalrus is good with this / the remaining comments are minor - any blockers to get this over the line?

@ChaoticWalrus, I just wanted to confirm we're aligned on the changes for the remaining unresolved items. Also if you see any hurdles to getting this in

ChaoticWalrus:
> @ChaoticWalrus, I just wanted to confirm we're aligned on the changes for the remaining unresolved items. Also if you see any hurdles to getting this in

Things are looking fine to me, overall.
Some more explanatory comments / noting somewhere about some of the finer points we discussed seems reasonable.

src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -0,0 +1,482 @@
...
+    /// @notice Validates that the cumulative stake of signed messages meets or exceeds the required threshold.
+    /// @param _signedWeight The cumulative weight of the signers that have signed the message.
+    /// @param _referenceBlock The block number to verify the stake threshold for
+    function _validateThresholdStake(uint256 _signedWeight, uint32 _referenceBlock) internal view {
+        uint256 totalWeight = _getTotalWeight(_referenceBlock);
+        if (_signedWeight > totalWeight){
+            revert InvalidSignedWeight();
+        }
+        uint256 thresholdStake = _getThresholdStake(_referenceBlock);
+        if ((thresholdStake * BPS)/ totalWeight > (_signedWeight * BPS) / totalWeight){
```
ChaoticWalrus: looks like you actually totally changed the meaning of `thresholdStake` to be the literal amount required to sign?
this check could now just be `if( thresholdStake > _signedWeight )` because you're doing the exact same manipulations on both sides of the inequality

ChaoticWalrus (approved):
I'm OK merging this into the 'unaudited' folder at this point, personally.
would be good to note any outstanding discussions, probably with some comments?
also might be nice to run this file structure by someone else, but I think it's quite clear having an "unaudited" folder 🤷

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Addressed in [79d95aa](https://github.com/Layr-Labs/eigenlayer-middleware/pull/194/commits/79d95aaf40f7b27f85ced4c42d2fce2d2086a865)

PR #202, opened 2024-03-08
feat: nonsigning rate helpers

0x0aa0:
- Adds function getQuorumBitmapsAtBlockNumber() to retrieve a set bitmaps at a block number given a set of operatorIds
- Adds operatorId to events OperatorAddedToQuorums() and OperatorRemovedFromQuorums()


src/OperatorStateRetriever.sol
```diff
@@ -159,4 +159,23 @@ contract OperatorStateRetriever {
...
         return checkSignaturesIndices;
     }
+
+    /**
+     * @notice this function returns the quorumBitmaps for each of the operators in the operatorIds array at the given blocknumber
+     * @param registryCoordinator is the AVS registry coordinator to fetch the operator information from
+     * @param operatorIds are the ids of the operators to get the quorumBitmaps for
+     * @param blockNumber is the block number to get the quorumBitmaps for
+     */
+    function getQuorumBitmapsAtBlockNumber(
```
jianoaix, line 169: I wonder what's the failure cases for this function, like:
- for an operatorID, in what cases it'll fail to get bitmap? Do we distinguish between not found v.s. some real error?
- does this function require all operatorIDs to find bitmap to succeed ("ALL-OF" v.s. "ONE-OF")? If it's ONE-OF semantics, what will be included in quorumBitmaps[i] if operatorIds[i] failed to get result?

jianoaix (commented):
Thank you Arch, this will reduce the RPCs needed from `N+1` to just 1 (at this offchain code: https://github.com/Layr-Labs/eigenda/blob/master/core/eth/tx.go#L577)

src/OperatorStateRetriever.sol
ChaoticWalrus: these are excellent questions to be asking
AFAICT if it doesn't find a value for a single operator, it should revert entirely.

test/events/IBLSApkRegistryEvents.sol
```diff
@@ -11,12 +11,14 @@ interface IBLSApkRegistryEvents {
...
     event OperatorAddedToQuorums(
         address operator,
+        bytes32 operatorId,
         bytes quorumNumbers
     );
 
     // @notice Emitted when an operator pubkey is removed from a set of quorums
     event OperatorRemovedFromQuorums(
         address operator, 
+        bytes32 operatorId,
```
ChaoticWalrus, line 21: seems like a good change; need to watch out for this breaking subgraphs etc

ChaoticWalrus (approved):
LGTM

PR #204, opened 2024-03-11
Reduce gas in getApkIndicesAtBlockNumber

uri-lightblocks:
feat: reduce gas cost 
with @RonTuretzky


RonTuretzky:
quorumApkUpdatesLength cannot be 0 due to the require on line 213

PR #205, opened 2024-03-11
Fix metadata uri signature

RonTuretzky:
Renaming for consistency with AVS Standard V1 Marketplace guidelines 

@ypatil12 🙏🏻


ypatil12 (approved):
Thanks for this PR, approving :)

PR #209, opened 2024-03-13
feat: add experimental contracts

ChaoticWalrus:
ripped the best contracts from https://github.com/Layr-Labs/eigenlayer-middleware/pull/165

this is assorted quorum-related logic, plus epoch-based registries (that use ECDSA sigs for convenience / ease of writing things)

### DIFF
--- a/src/unaudited/EcdsaEpochRegistry_Modular.sol
+++ b/src/unaudited/EcdsaEpochRegistry_Modular.sol
@@ -1,0 +1,195 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity =0.8.12;
+
+import {IStrategy} from "eigenlayer-contracts/src/contracts/interfaces/IStrategy.sol";
+import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
+import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
+import {IServiceManager} from "../interfaces/IServiceManager.sol";
+import {LinearWeightQuorum} from "./LinearWeightQuorum.sol";
+
+import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+
+/** 
+ * @notice ~Equivalent to the `EcdsaEpochRegistry` contract, but built from modular pieces instead of
+ * a more monolithic construction.
+ */
+contract EcdsaEpochRegistry_Modular is LinearWeightQuorum {
+    /// @notice the ServiceManager for this AVS, which forwards calls onto EigenLayer's core contracts
+    IServiceManager public immutable serviceManager;
+
+    // @notice the length of each epoch, in seconds
+    uint256 public immutable epochLengthSeconds;
+    // @notice the start of the initial epoch
+    uint256 public immutable epochZeroStart;
+
+    /// @notice History of the total stakes
+    /// Mapping: epoch => total stake in epoch
+    mapping(uint256 => uint256) public totalStakeHistory;
+
+    // TODO: possibly pack the first 2 variables here
+    /// @notice Mapping: operator address => epoch => stake of operator in epoch
+    mapping(address => mapping(uint256 => uint256)) internal _operatorStakeHistory;
+
+    // TODO: this corresponds with suggested packing, above.
+    function operatorStakeHistory(address operator, uint256 epoch) public view returns (uint256) {
+        return _operatorStakeHistory[operator][epoch];
+    }
+
+    enum OperatorStatus {
+        // default is NEVER_REGISTERED
+        NEVER_REGISTERED,
+        REGISTERED,
+        DEREGISTERED
+    }
+
+    // TODO: documentation
+    mapping(address => OperatorStatus) operatorStatus;
+
+    // @notice: mapping: epoch => set of operators for the epoch
+    mapping(uint256 => address[]) public operatorsForEpoch;
+    // TODO: could probably restructure `operatorStakeHistory` to also jam that into the same slot as above
+
+    // storage gap for upgradeability
+    // slither-disable-next-line shadowing-state
+    uint256[46] private __GAP;
+
+    /** 
+     * @dev setting the `_epochZeroStart` to be in the past is disallowed.
+     * If you try to do so, the current block.timestamp will be used instead.
+     * Future times are allowed, with no safety check in place.
+     */
+    constructor(
+        IServiceManager _serviceManager,
+        IDelegationManager _delegationManager,
+        uint256 _epochLengthSeconds,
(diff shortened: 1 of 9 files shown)


src/unaudited/EcdsaEpochRegistry_Permissionless.sol
```diff
@@ -0,0 +1,250 @@
...
+            // add any registered operator that meets the weight requirement to the current epoch's operator set
+            if (operatorWeight >= minimumWeightRequirement) {
+                operatorsForEpoch[_currentEpoch].push(operator);
+                _operatorStakeHistory[operator][_currentEpoch] = operatorWeight;
+                // TODO: add event
+                if (operatorWeight < _realizedMinimumWeight) {
+                    _realizedMinimumWeight = operatorWeight;
+                }
+            // otherwise, evict the operator
+            } else {
```
stevennevins, line 129: It seems unnecessary to kick the operators here.

Then, they may be considered in a new epoch without any additional action, either by retargeting, deregistration of other operators, or new delegations. They might meet the minimum requirements.  

This could complicate payments?

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
```diff
@@ -0,0 +1,250 @@
...
+            operatorIndex[operator] = numberRegisteredOperators();
+            registeredOperators.push(operator);
+        }
+    }
+
+    // TODO: documentation
+    function _deregisterOperator(
+        address operator
+    ) internal virtual override {
+        if (operatorStatus[operator] == OperatorStatus.REGISTERED) {
```
stevennevins, line 171: What is the intended behavior for an operator deregistering when serving in an epoch?  Are they still expected to uphold their commitment to the AVS until the epoch ends?

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
```diff
@@ -0,0 +1,250 @@
...
+            uint256 previousWeightRequirement = minimumWeightRequirement;
+            uint256 _targetOperatorSetSize = targetOperatorSetSize;
+
+            // start with "adjustment" of 1 (i.e. leaving the `minimumWeightRequirement` value unchanged)
+            uint256 adjustmentFactor = 1e18;
+            if (currentOperatorSetSize >= _targetOperatorSetSize) {
+                // when current = target, no change
+                // when current = max, add the retargetingFactorWei
+                // when current is halfway between target and max, add half of retargetingFactorWei
+                adjustmentFactor += retargetingFactorWei * (currentOperatorSetSize - _targetOperatorSetSize) / (maxRegisteredOperators - _targetOperatorSetSize);
```
stevennevins, line 208: The retargeting function is linear afaict

If this was based on the regular gas market EIP1559, it's a linear retargeting.  The blob space market 1559 is the only one that uses exponential retargeting, which has improved the smoothing of demand spikes and is the one we should base this off

src/unaudited/QuorumLib.sol
```diff
@@ -0,0 +1,73 @@
...
+    struct Quorum {
+        IStrategy[] strategies;
+        uint256[] weights;
+    }
+
+    /**
+     * @notice This function computes the total weight of the @param operator.
+     * @return `uint256` The weighted sum of the operator's shares across each strategy considered
+     */
+    function linearWeightOfOperator(
```
stevennevins, line 37: You could generalize this library to take a function pointer from LinearWeightQuorum.sol that performs the linear calculation.  That way this is a QuorumLib vs LinearQuorumLib

src/unaudited/EcdsaEpochRegistry_Modular.sol
```diff
@@ -0,0 +1,195 @@
...
+            operatorStatus[operator] = OperatorStatus.DEREGISTERED;
+            serviceManager.deregisterOperatorFromAVS(operator);
+
+            // TODO: event
+            // emit OperatorDeregistered(operator, operatorId);
+        }
+    }
+
+    // @notice TODO: proper documentation
+    function _evaluateOperatorSet(
```
stevennevins, line 145: Wdyt, about a where where operators re-confirm their status as an operator for the upcoming epoch instead of the design where we do updates for the entire operator set. I think this epoch design + operators having a `poke` like function to signal they want to join the next epoch could alleviate some of the updateStakes gas cost issues

ChaoticWalrus: I think that's OK; as a model in general, requiring a timely `poke` kind of operation from operators seems reasonable.
idk if it's really much better though. kind of just shifts some cost + responsibility to operators, at least in my mind.

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
ChaoticWalrus: hmm, good point that if the target goes down, you might not want to kick people before then.
I believe my thought process was "we are checking if they meet the current requirement, then we will retarget".
will give the optimal design more thought though.

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
ChaoticWalrus: yes. they would be expected to continue serving the AVS until the epoch ends.

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
ChaoticWalrus: hmm, maybe I misunderstood the way these terms are being used.
I was viewing this as exponential retargeting since the net effect across epochs is the weight requirement being repeatedly multiplied by a factor (a la exponentiation), but I guess you're saying it would be exponential if the adjustment factor calculated was
`retargetingFactorWei` _raised to a power_ dependent on the difference between `currentOperatorSetSize` and `_targetOperatorSetSize`
instead of what is there now.
I think this is super reasonable; I specifically didn't try to optimize the algorithm here at all.

src/unaudited/QuorumLib.sol
ChaoticWalrus: big brain idea right there.
I would want to do gas testing on this to see if any of this is even worthwhile, tbh.
I feel like this "should" all be a library, but obviously it's not worth it if it makes the calculations appreciably more expensive.

src/unaudited/EcdsaEpochRegistry_Modular.sol
stevennevins: socializing the cost and responsibility seems beneficial

src/unaudited/EcdsaEpochRegistry_Permissionless.sol
stevennevins: makes sense, the latter seems more in line with what we want.  https://dankradfeist.de/ethereum/2022/03/16/exponential-eip1559.html

PR #215, opened 2024-03-21
feat: zk stake updates

stevennevins:


src/RegistryCoordinator.sol
```diff
@@ -342,6 +342,63 @@ contract RegistryCoordinator is
...
+            return (operators, quorum, stakes);
+    }
+
+    function zkUpdateOperatorsForQuorum(uint256 updateBlock, address[] memory operators, uint8 quorum, uint96[] memory stakes, uint256 totalStake, bytes32 postStateDigest, bytes calldata seal) external {
+
+        uint256 lastQuorumUpdateBlock = quorumUpdateBlockNumber[quorum];
+        require(lastQuorumUpdateBlock < updateBlock, "Update is Stale");
+        /// Check that the current update is before the lastQuorumUpdate
+
+        bytes memory journal = abi.encode(operators, quorum, stakes, totalStake);
```
gpsanant: hmm when was the view function called, perhaps i need to look at the guest

PR #216, opened 2024-03-26
chore: less restrictive pragma

stevennevins:


samlaf:
This would be super useful. For eg integrating our contracts with risc0 contracts was a pain because they had their contracts using ^0.8.21

RonTuretzky:
bumping for hype 🔥

wadealexc (changes requested):
Just needs a submodule update for the core repo, and then this is good to go!

wadealexc (approved):
(pending submodule update, LGTM otherwise)

PR #227, opened 2024-04-12
chore: bump forge-std and remove ds-test dependency

stevennevins:


samlaf (approved):
Oh awesome this completely removes the dependency on ds-test, and we can now use forge-std 1.8.1 @dabit3 :)

ChaoticWalrus:
LGTM, merging.

PR #228, opened 2024-04-16
feat: return operators ejected

0x0aa0:
ejectOperators() returns the number of operators actually ejected for each quorum

### DIFF
--- a/src/interfaces/IEjectionManager.sol
+++ b/src/interfaces/IEjectionManager.sol
@@ -32,7 +32,7 @@ interface IEjectionManager {
      * @notice Ejects operators from the AVSs registryCoordinator under a ratelimit
      * @param _operatorIds The ids of the operators to eject for each quorum
      */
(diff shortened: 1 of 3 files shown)


gpsanant (approved):

jianoaix (approved):
Thanks

src/interfaces/IEjectionManager.sol
```diff
@@ -32,7 +32,7 @@ interface IEjectionManager {
      * @notice Ejects operators from the AVSs registryCoordinator under a ratelimit
      * @param _operatorIds The ids of the operators to eject for each quorum
      */
-    function ejectOperators(bytes32[][] memory _operatorIds) external;
+    function ejectOperators(bytes32[][] memory _operatorIds) external returns (uint256[] memory);
```
jianoaix, line 35: Sorry for late feedback: can we return uint32 instead? There won't be so many operators to eject

PR #234, opened 2024-04-18
feat: ecdsa service manager

stevennevins:


src/unaudited/ECDSAServiceManagerBase.sol
```diff
@@ -0,0 +1,240 @@
...
+                address(this),
+                rangePayments[i].amount
+            );
+            rangePayments[i].token.approve(
+                paymentCoordinator,
+                rangePayments[i].amount
+            );
+        }
+
+        IPaymentCoordinator(paymentCoordinator).payForRange(rangePayments);
```
ChaoticWalrus, line 177: only 'gotcha' I can think of is if there are duplicate tokens, this will revert at the end due to insufficient allowance.
I'm fine chalking that up as "operator error" though.

ChaoticWalrus (approved):
Seems very reasonable, looks clean, nice work 👍

src/unaudited/ECDSAServiceManagerBase.sol
yorhodes, line 1: I dont really understand how any of this is ECDSA specific

stevennevins: The naming was to signal that it's associated with the ECDSAStakeRegistry

PR #240, opened 2024-04-19
feat: reregistration delay and BLS key rotation

0x0aa0:
Allows operators to rotate BLS keys by fully deregistering and reregistering with a new signature

### DIFF
--- a/src/BLSApkRegistryStorage.sol
+++ b/src/BLSApkRegistryStorage.sol
@@ -29,6 +29,9 @@ abstract contract BLSApkRegistryStorage is Initializable, IBLSApkRegistry {
     /// @notice maps quorumNumber => current aggregate pubkey of quorum
     mapping(uint8 => BN254.G1Point) public currentApk;
 
+    /// @notice maps operator address to pubkey history
+    mapping(address => PubkeyCheckpoint[]) public operatorPubkeyHistory;
+
     constructor(IRegistryCoordinator _registryCoordinator) {
         registryCoordinator = address(_registryCoordinator);
         // disable initializers so that the implementation contract cannot be initialized
@@ -36,5 +39,5 @@ abstract contract BLSApkRegistryStorage is Initializable, IBLSApkRegistry {
     }
 
     // storage gap for upgradeability
-    uint256[45] private __GAP;
+    uint256[44] private __GAP;
 }
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -507,6 +507,12 @@ contract RegistryCoordinator is
         if (operatorId == 0) {
             operatorId = blsApkRegistry.registerBLSPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
         }
+        if (_operatorInfo[operator].status == OperatorStatus.DEREGISTERED && 
+            params.pubkeyRegistrationSignature.X != 0 &&
+            params.pubkeyRegistrationSignature.Y != 0
+        ) {
+            operatorId = blsApkRegistry.registerBLSPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
+        }
         return operatorId;
     }
 
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -511,7 +511,8 @@ contract RegistryCoordinator is
         }
         if (_operatorInfo[operator].status == OperatorStatus.DEREGISTERED && 
             params.pubkeyRegistrationSignature.X != 0 &&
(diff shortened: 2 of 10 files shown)


src/RegistryCoordinator.sol
```diff
@@ -471,6 +471,8 @@ contract RegistryCoordinator is
         // If the operator wasn't registered for any quorums, update their status
         // and register them with this AVS in EigenLayer core (DelegationManager)
         if (_operatorInfo[operator].status != OperatorStatus.REGISTERED) {
+            require(lastDeregistrationTimestamp[operator] + REREGISTRATION_DELAY < block.timestamp, "RegistryCoordinator._registerOperator: operator cannot reregister yet");
```
stevennevins, line 474: Can we have an `updateBLSPublickey` that is outside of the registration flow?  we should avoid the 7 day re-registration delay if possible

src/RegistryCoordinator.sol
```diff
@@ -507,6 +509,12 @@ contract RegistryCoordinator is
         if (operatorId == 0) {
             operatorId = blsApkRegistry.registerBLSPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
         }
+        if (_operatorInfo[operator].status == OperatorStatus.DEREGISTERED && 
+            params.pubkeyRegistrationSignature.X != 0 &&
+            params.pubkeyRegistrationSignature.Y != 0 
+        ) {
+            operatorId = blsApkRegistry.registerBLSPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
```
ChaoticWalrus, line 516: won't this call return the hash of the _new_ pubkey? might have to leave the local `operatorId` variable unchanged / explicitly ignore the return value here?

perhaps I'm misremembering something, but I believe the return value of _this_ function (`_getOrCreateOperatorId`) is used in other functions after this is called. 

Or perhaps I'm misinterpreting, and the operator's ID is actually changing here + supposed to be changing?

I guess let me know the intended/desired behavior here, because I'm not quite sure what it is

src/interfaces/IBLSApkRegistry.sol
```diff
@@ -33,6 +33,18 @@ interface IBLSApkRegistry is IRegistry {
         BN254.G2Point pubkeyG2;
     }
 
+    /**
+     * @notice Struct used to checkpoint the previous pubkeys and their hashes for a given operator
+     * @param previousPubkeyG1 is the G1 public key of the operator
+     * @param previoudPubkeyHash is the hash of the public key of the operator
```
ChaoticWalrus, line 39: typo: 'previoud' instead of 'previous'

src/BLSApkRegistryStorage.sol
```diff
@@ -29,12 +29,15 @@ abstract contract BLSApkRegistryStorage is Initializable, IBLSApkRegistry {
     /// @notice maps quorumNumber => current aggregate pubkey of quorum
     mapping(uint8 => BN254.G1Point) public currentApk;
 
+    /// @notice maps operator address to pubkey history
+    mapping(address => PubkeyCheckpoint[]) public operatorPubkeyHistory;
```
ChaoticWalrus, line 33: might be worth noting (maybe an @ dev tag?) that this excludes the _current_ key? feels like something that might be easy to misinterpret, at least?

src/RegistryCoordinator.sol
ChaoticWalrus: Seems this is indeed changing the operator's ID, so this behavior seems appropriate, assuming changing the ID is the intention 👍

src/RegistryCoordinator.sol
```diff
@@ -507,6 +509,12 @@ contract RegistryCoordinator is
         if (operatorId == 0) {
             operatorId = blsApkRegistry.registerBLSPublicKey(operator, params, pubkeyRegistrationMessageHash(operator));
         }
+        if (_operatorInfo[operator].status == OperatorStatus.DEREGISTERED &&
```
ChaoticWalrus, line 512: probably should be an `else if` rather than just an `if`, even if the cases are (supposed to be) mutually exclusive?

ChaoticWalrus (commented):
Seems quite good! I have a couple questions + suggestions, which I've noted.

Broadly, it looks like this is actually changing the operator's ID, in addition to their BLS key -- so it is _keeping_ the assumption that
operatorID = hash of operator's key
but breaking the previous behavior of operators having unchanging IDs.

Is there any storage we have that is keyed to Operator IDs rather than operator addresses? This seems very well-contained at the moment, but I would be concerned about understanding the implications of this change if such storage does exist. If it doesn't exist, then great!

PR #242, opened 2024-04-22
feat: service manager payments

8sunyuan:
Adds the function `submitRangePayments` to the ServiceManagerBase contract which allows the ServiceManager owner
to submit RangePayments to the PaymentCoordinator. Note that the owner must have already approved the ServiceManager contract for the payment tokens they are submitting.
Includes interface changes as well to be more explicit on the UI required interfaces.

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -6,6 +6,7 @@ import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/Ownabl
 import {BitmapUtils} from "./libraries/BitmapUtils.sol"; 
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import {IAVSDirectory} from "eigenlayer-contracts/src/contracts/interfaces/IAVSDirectory.sol";
+import {IPaymentCoordinator} from "eigenlayer-contracts/src/contracts/interfaces/IPaymentCoordinator.sol";
 
 import {IServiceManager} from "./interfaces/IServiceManager.sol";
 import {IRegistryCoordinator} from "./interfaces/IRegistryCoordinator.sol";
@@ -22,6 +23,7 @@ abstract contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
     IRegistryCoordinator internal immutable _registryCoordinator;
     IStakeRegistry internal immutable _stakeRegistry;
     IAVSDirectory internal immutable _avsDirectory;
+    IPaymentCoordinator internal immutable _paymentCoordinator;
 
     /// @notice when applied to a function, only allows the RegistryCoordinator to call it
     modifier onlyRegistryCoordinator() {
@@ -36,11 +38,13 @@ abstract contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
     constructor(
         IAVSDirectory __avsDirectory,
         IRegistryCoordinator __registryCoordinator,
-        IStakeRegistry __stakeRegistry
+        IStakeRegistry __stakeRegistry,
+        IPaymentCoordinator ___paymentCoordinator
     ) {
         _avsDirectory = __avsDirectory;
         _registryCoordinator = __registryCoordinator;
         _stakeRegistry = __stakeRegistry;
+        _paymentCoordinator = ___paymentCoordinator;
         _disableInitializers();
     }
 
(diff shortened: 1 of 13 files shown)


src/ServiceManagerBase.sol
```diff
@@ -36,11 +39,13 @@ abstract contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
     constructor(
         IAVSDirectory __avsDirectory,
         IRegistryCoordinator __registryCoordinator,
-        IStakeRegistry __stakeRegistry
+        IStakeRegistry __stakeRegistry,
+        IPaymentCoordinator ___paymentCoordinator
```
samlaf: is IPaymentCoordinator an eigenlayer core contrac? if so, can I be autistic and ask to move it above registryCoordinator, to separate the core contracts from middleware contracts (just easier to think I find..)

src/ServiceManagerBase.sol
```diff
@@ -57,6 +62,32 @@ abstract contract ServiceManagerBase is IServiceManager, OwnableUpgradeable {
...
+     * and of course has the required balances.
+     * @param rangePayments The range payments being created
+     * @dev Expected to be called by the ServiceManager of the AVS on behalf of which the payment is being made
+     * @dev The duration of the `rangePayment` cannot exceed `paymentCoordinator.MAX_PAYMENT_DURATION()`
+     * @dev The tokens are sent to the `PaymentCoordinator` contract
+     * @dev Strategies must be in ascending order of addresses to check for duplicates
+     * @dev This function will revert if the `rangePayment` is malformed,
+     * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     */
+    function submitRangePayments(
```
samlaf: curious, is there a reason we are not calling this function payForRange?
Do we have a policy of reusing same name function only when proxying the exact same call, whereas here we are also transferring tokens and approving?

src/ServiceManagerBase.sol
```diff
@@ -6,8 +6,10 @@ import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/Ownabl
 import {BitmapUtils} from "./libraries/BitmapUtils.sol"; 
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
 import {IAVSDirectory} from "eigenlayer-contracts/src/contracts/interfaces/IAVSDirectory.sol";
+import {IPaymentCoordinator} from "eigenlayer-contracts/src/contracts/interfaces/IPaymentCoordinator.sol";
 
 import {IServiceManager} from "./interfaces/IServiceManager.sol";
+// import {IServiceManagerUI} from "./interfaces/IServiceManagerUI.sol";
```
samlaf: delete

src/interfaces/IServiceManagerUI.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.5.0;
+
+import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
+import {IDelegationManager} from "eigenlayer-contracts/src/contracts/interfaces/IDelegationManager.sol";
+
+/**
+ * @title Minimal interface for a ServiceManager-type contract that AVS ServiceManager contracts must implement
+ * to index their data for the V1 AVS Marketplace
```
samlaf: can we change this to: "[...] must implement for eigenlabs to be able to index their data [...]"? Otherwise sounds like this indexing is something they would want to do themselves and the AVS teams might get confused.

samlaf (commented):
Thanks for separating IServiceManagerUI into its own separate interface. I only reviewed the src contracts (not tests), lgtm! Added some small nitpick comments.

src/ServiceManagerBase.sol
8sunyuan: That was my original thought to rename it since it has some additional ERC20 token logic but its essentially proxying the same call. I renamed to `payForRange` for consistency sake with the AVSDirectory calls having same naming.

https://github.com/Layr-Labs/eigenlayer-middleware/pull/242/commits/cbe2107dd590fd49c5aa97274ca7beecb9b19644

src/ServiceManagerBase.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/242/commits/cbe2107dd590fd49c5aa97274ca7beecb9b19644

src/ServiceManagerBase.sol
8sunyuan: Done
https://github.com/Layr-Labs/eigenlayer-middleware/pull/242/commits/cbe2107dd590fd49c5aa97274ca7beecb9b19644

src/interfaces/IServiceManagerUI.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/242/commits/cbe2107dd590fd49c5aa97274ca7beecb9b19644
Changed to 
```
 * @title Minimal interface for a ServiceManager-type contract that AVS ServiceManager contracts must implement
 * for eigenlabs to be able to index their data on the AVS marketplace frontend.
```

samlaf (approved):
Thanks for quick changes. LGTM now. I'll approve even though 2 tests are failing and should be fixed, and someone else should review the tests.

PR #246, opened 2024-04-26
feat: reregistration delay

0x0aa0:
Adds a delay for registration to any quorum after an operator has been ejected

### DIFF
--- a/src/EjectionManager.sol
+++ b/src/EjectionManager.sol
@@ -60,14 +60,11 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
     /**
      * @notice Ejects operators from the AVSs RegistryCoordinator under a ratelimit
      * @param _operatorIds The ids of the operators 'j' to eject for each quorum 'i'
-     * @return ejectedOperatorsForQuorum The total number of operators ejected for each quorum 'i'
-     * @dev This function will eject as many operators as possible without reverting prioritizing operators at the lower index
+     * @dev This function will eject as many operators as possible prioritizing operators at the lower index
      * @dev The owner can eject operators without recording of stake ejection
      */
-    function ejectOperators(bytes32[][] memory _operatorIds) external returns (uint32[] memory){
+    function ejectOperators(bytes32[][] memory _operatorIds) external {
         require(isEjector[msg.sender] || msg.sender == owner(), "Ejector: Only owner or ejector can eject");
-
-        uint32[] memory ejectedOperatorsForQuorum = new uint32[](_operatorIds.length);
 
         for(uint i = 0; i < _operatorIds.length; ++i) {
             uint8 quorumNumber = uint8(i);
@@ -76,7 +73,7 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
             uint256 stakeForEjection;
             uint32 ejectedOperators;
 
-            bool broke;
+            bool ratelimitHit;
             for(uint8 j = 0; j < _operatorIds[i].length; ++j) {
                 uint256 operatorStake = stakeRegistry.getCurrentStake(_operatorIds[i][j], quorumNumber);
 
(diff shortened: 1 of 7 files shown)


src/RegistryCoordinator.sol
```diff
@@ -423,6 +424,16 @@ contract RegistryCoordinator is
...
     }
 
+    /**
+     * @notice Sets the reregistration delay, which is the time an operator must wait after 
+     * ejection before registering for any quorum
+     * @param _reregistrationDelay the new reregistration delay
+     * @dev only callable by the owner
+     */
+    function setReregistrationDelay(uint256 _reregistrationDelay) external onlyOwner {
+        reregistrationDelay = _reregistrationDelay;
```
ChaoticWalrus: might want to make the name of this reflect the fact that it's only used when ejection occurs?
e.g. the 'reregistration delay' (current name) is not enforced when an operator is 'churned out'

as a side note, I think this may present a minor "attack" whereby an operator de-registers themself instead of getting ejected, to evade this delay?
You could address this by making `ejectOperator` always set the `lastEjectionTimestamp` and no longer revert if the operator is not registered.

ChaoticWalrus: suggested new names:
ejectionLength
ejectionReregistrationDelay or ejectionReregistrationPeriod
ejectionCooldownPeriod

I don't particularly like any of these but these are some ideas

src/RegistryCoordinatorStorage.sol
```diff
@@ -64,6 +64,11 @@ abstract contract RegistryCoordinatorStorage is IRegistryCoordinator {
     /// @notice the address of the entity allowed to eject operators from the AVS
     address public ejector;
 
+    /// @notice the last timestamp an operator was ejected
+    mapping(address => uint256) public lastEjectionTimestamp;
+    /// @notice the delay before an operator can reregister after being ejected
```
ChaoticWalrus: should probably do a dev tag + explain this is in seconds? or otherwise have seconds in the name (but this may be 'gross' / lead to a sentence-long name)

ChaoticWalrus (changes requested):
I believe there's a ~medium-severity issue here -- see my comment.
I would definitely like to resolve this discussion before merging.

src/RegistryCoordinator.sol
```diff
@@ -364,10 +364,22 @@ contract RegistryCoordinator is
...
+
+        OperatorInfo storage operatorInfo = _operatorInfo[operator];
+        bytes32 operatorId = operatorInfo.operatorId;
+        uint192 quorumsToRemove = uint192(BitmapUtils.orderedBytesArrayToBitmap(quorumNumbers, quorumCount));
+        uint192 currentBitmap = _currentOperatorBitmap(operatorId);
+        if(
+            operatorInfo.status == OperatorStatus.REGISTERED && 
+            !quorumsToRemove.isEmpty() &&
+            quorumsToRemove.isSubsetOf(currentBitmap) 
+        ){
```
ChaoticWalrus, line 378: so there's still some weird race condition where an operator can deregister from a subset of quorums to avoid being ejected for more, right?
I guess I can accept that, but it would certainly be nice to resolve if there's an easy solution.

ChaoticWalrus: perhaps we can just take the submitted 'quorumNumbers' and actually pass on a combination ('and'ing) of them with the operator's current quorums in the subsequent call?

ChaoticWalrus:
I could see merging this now. Am pushing a bit on fixing the remaining race condition here, at least if it's easy enough, because it would be a nice improvement.
Otherwise LGTM

ChaoticWalrus (approved):

src/EjectionManager.sol
```diff
@@ -90,35 +87,31 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
...
                         stakeEjected: stakeForEjection
                     }));
-                    broke = true;
+                    ratelimitHit = true;
                     break;
                 }
 
-                //try-catch used to prevent race condition of operator deregistering before ejection
-                try registryCoordinator.ejectOperator(
+                stakeForEjection += operatorStake;
```
stevennevins, line 94: When the rate limit is hit, will we not update these state variables or eject the last operator who has their stake pushed onto the stakeEjectedForQuorum?

stevennevins: due to the `break`*

stevennevins: Ahh nvm this looks good.  The variables are memory, and we don't need to eject that operator because we crossed the rate limit

src/EjectionManager.sol
```diff
@@ -60,23 +60,20 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
     /**
      * @notice Ejects operators from the AVSs RegistryCoordinator under a ratelimit
      * @param _operatorIds The ids of the operators 'j' to eject for each quorum 'i'
-     * @return ejectedOperatorsForQuorum The total number of operators ejected for each quorum 'i'
-     * @dev This function will eject as many operators as possible without reverting prioritizing operators at the lower index
+     * @dev This function will eject as many operators as possible prioritizing operators at the lower index
      * @dev The owner can eject operators without recording of stake ejection
      */
-    function ejectOperators(bytes32[][] memory _operatorIds) external returns (uint32[] memory){
+    function ejectOperators(bytes32[][] memory _operatorIds) external {
```
stevennevins, line 66: wdyt about refactoring the ejection logic 

1. That doesn't enforce the rate limit that is callable by the owner
2. That is callable by the ejector and enforces the rate limit

This should clean up some of the branchy conditionals in this function

src/EjectionManager.sol
stevennevins: A high level comment, I don't love logic occurring after `break` in a loop and all the conditional logic to enter the break condition. A `while` loop is probably more readable here

stevennevins (approved):

PR #248, opened 2024-04-29
feat: batch operator id conversions

0x0aa0:
Adds two view functions to OperatorStateRetriever:
- getBatchOperatorId: batch conversion of operator address to operator id
- getBatchOperatorFromId: batch conversion of operator id to operator address

### DIFF
--- a/src/OperatorStateRetriever.sol
+++ b/src/OperatorStateRetriever.sol
@@ -179,6 +179,12 @@ contract OperatorStateRetriever {
         return quorumBitmaps;
     }
 
+    /**
+     * @notice This function returns the operatorIds for each of the operators in the operators array
+     * @param registryCoordinator is the AVS registry coordinator to fetch the operator information from
+     * @param operators is the array of operator address to get corresponding operatorIds for
+     * @dev if an operator is not registered, the operatorId will be 0
+     */
     function getBatchOperatorId(
         IRegistryCoordinator registryCoordinator,
         address[] memory operators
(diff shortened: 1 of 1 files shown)


src/OperatorStateRetriever.sol
```diff
@@ -178,4 +178,25 @@ contract OperatorStateRetriever {
         }
         return quorumBitmaps;
     }
+
+    function getBatchOperatorId(
```
jianoaix, line 188: These two methods will revert if any of the ID/address doesn't exist?

jianoaix (commented):
LG

src/OperatorStateRetriever.sol
ChaoticWalrus: as written, these methods currently will _not_ revert, but instead return `bytes32(0)` as the ID of any operator who has never registered, or `address(0)` for any operatorID that has never been used.

relevant code:
https://github.com/Layr-Labs/eigenlayer-middleware/blob/feat/batch-operatorid-view/src/BLSApkRegistry.sol#L275-L276
https://github.com/Layr-Labs/eigenlayer-middleware/blob/feat/batch-operatorid-view/src/RegistryCoordinator.sol#L796-L804

@jianoaix what would your preferred behavior here be?

ChaoticWalrus (changes requested):
Looks fine. Would be good to resolve the thread around reversions + add some documentation, prior to merging.

src/OperatorStateRetriever.sol
jianoaix: Yep, that's more desirable! Could we put this in the comment of the function?

0x0aa0: No they will just return a 0 value at that index

jianoaix (commented):
LGTM

ChaoticWalrus (approved):
LGTM.
FYI, the 2 automated test failures are unreleated to this PR and are getting fixed in a separate PR.

PR #249, opened 2024-04-29
feat: BLS key rotation

0x0aa0:


src/RegistryCoordinator.sol
```diff
@@ -480,8 +490,12 @@ contract RegistryCoordinator is
         require(quorumsToAdd.noBitsInCommon(currentBitmap), "RegistryCoordinator._registerOperator: operator already registered for some quorums being registered for");
         uint192 newBitmap = uint192(currentBitmap.plus(quorumsToAdd));
 
-        // Check that the operator can reregister if ejected
-        require(lastEjectionTimestamp[operator] + ejectionCooldown < block.timestamp, "RegistryCoordinator._registerOperator: operator cannot reregister yet");
+        // Check that the operator can reregister 
+        require(
+            lastEjectionTimestamp[operator] + ejectionCooldown < block.timestamp &&
+            lastDeregistrationTimestamp[operator] + deregistrationCooldown < block.timestamp,
```
samlaf, line 496: why do we need a deregistrationCooldown if we keep the entire history of keys onchain?

samlaf: Or is 0 a reasonable value for deregistrationCooldown?

samlaf (commented):
Just did a quick glance review. Will try to summarize what this is doing to make sure I'm not missing anything:
- operator can either be ejected (force deregister) or deregister by themself
- both of those have cooldowns (ejectionCooldown and deregistrationCooldown) which force the operator to wait before reregistering
- We keep the entire history of registered bn254 G1 pubkey hash onchain, such that the AVS can slash an operator for misbehavior using a previous key

Have a Q related to why we need deregistrationCooldown below.

src/RegistryCoordinator.sol
0x0aa0: @gpsanant has suggested this should be a few days. I think something like an hour is more reasonable

src/BLSApkRegistry.sol
```diff
@@ -95,21 +97,19 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
...
      * @param params contains the G1 & G2 public keys of the operator, and a signature proving their ownership
-     * @param pubkeyRegistrationMessageHash is a hash that the operator must sign to prove key ownership
      */
     function registerBLSPublicKey(
         address operator,
-        PubkeyRegistrationParams calldata params,
-        BN254.G1Point calldata pubkeyRegistrationMessageHash
+        PubkeyRegistrationParams calldata params
     ) external onlyRegistryCoordinator returns (bytes32 operatorId) {
+        BN254.G1Point memory _pubkeyRegistrationMessageHash = pubkeyRegistrationMessageHash(operator);
```
ChaoticWalrus: you could move this below the shortcircuit 'if' statement below, for a bit more efficiency

src/BLSApkRegistry.sol
```diff
@@ -123,18 +123,27 @@ contract BLSApkRegistry is BLSApkRegistryStorage {
...
         // e(sigma + P * gamma, [-1]_2) = e(H(m) + [1]_1 * gamma, P') 
         require(BN254.pairing(
             params.pubkeyRegistrationSignature.plus(params.pubkeyG1.scalar_mul(gamma)),
             BN254.negGeneratorG2(),
-            pubkeyRegistrationMessageHash.plus(BN254.generatorG1().scalar_mul(gamma)),
+            _pubkeyRegistrationMessageHash.plus(BN254.generatorG1().scalar_mul(gamma)),
             params.pubkeyG2
         ), "BLSApkRegistry.registerBLSPublicKey: either the G1 signature is wrong, or G1 and G2 private key do not match");
 
+        //if keys already registered checkpoint the previous pubkey and pubkeyHash of the operator
```
ChaoticWalrus: I found this challenging to parse. I feel like there's an implied comma between `if keys already registered` and `checkpoint the previous pubkey and pubkeyHash of the operator`? I'd recommend just inserting the word `then` in between these two phrases to make it clearer

src/BLSApkRegistryStorage.sol
```diff
@@ -29,12 +31,15 @@ abstract contract BLSApkRegistryStorage is Initializable, IBLSApkRegistry {
     /// @notice maps quorumNumber => current aggregate pubkey of quorum
     mapping(uint8 => BN254.G1Point) public currentApk;
 
+    /// @notice maps operator address to pubkey history
+    mapping(address => PubkeyCheckpoint[]) public operatorPubkeyHistory;
```
ChaoticWalrus, line 35: should we have something like a `pubkeyAtBlock` function for querying this? at minimum just a getter for the length?

src/RegistryCoordinator.sol
```diff
@@ -446,6 +446,16 @@ contract RegistryCoordinator is
...
 
+    /**
+     * @notice Sets the deregistration cooldown, which is the time an operator must wait in 
+     * seconds after fully deregistering before registering for any quorum
+     * @param _deregistrationCooldown the new deregistration cooldown in seconds
+     * @dev only callable by the owner
+     */
+    function setDeregistrationCooldown(uint256 _deregistrationCooldown) external onlyOwner {
+        deregistrationCooldown = _deregistrationCooldown;
+    }
```
ChaoticWalrus, line 457: needs an event.
might want this added to the initializer as well.

I'm considering if any input sanitization makes sense here too -- should this have a min or max? is 0 an OK input? I feel like it's probably OK to not have this, given the owner-only nature of the function... I don't see any super "unsafe" values

src/RegistryCoordinator.sol
ChaoticWalrus: this...limits the rate at which operators can edit their keys. I don't think it's serving a serious purpose beyond maybe addressing some theoretical worry where an operator keeps changing their key? I guess if you need to search past keys it's potentially problematic if the growth rate of the historical storage isn't limited?

src/BLSApkRegistry.sol
```diff
@@ -1,13 +1,12 @@
...
 
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
 import {BLSApkRegistryStorage} from "./BLSApkRegistryStorage.sol";
-
 import {IRegistryCoordinator} from "./interfaces/IRegistryCoordinator.sol";
-
 import {BN254} from "./libraries/BN254.sol";
 
-contract BLSApkRegistry is BLSApkRegistryStorage {
+contract BLSApkRegistry is BLSApkRegistryStorage, EIP712 {
```
ChaoticWalrus, line 9: making a note that I should verify that the `EIP712` contract is storage-free, or else this affects storage layout. perhaps it's OK either way though.

src/RegistryCoordinator.sol
0x0aa0: Didn't add to init because of bytecode size but I will see if it can be squeezed in. I could see really long values being not great for the delay but will probably not add sanitization for the same bytecode size issue

src/interfaces/IBLSApkRegistry.sol
```diff
@@ -33,6 +33,18 @@ interface IBLSApkRegistry is IRegistry {
...
+    /**
+     * @notice Struct used to checkpoint the previous pubkeys and their hashes for a given operator
+     * @param previousPubkeyG1 is the G1 public key of the operator
+     * @param previoudPubkeyHash is the hash of the public key of the operator
+     * @param blockNumber is the block number at which the pubkey was updated
+     */
+    struct PubkeyCheckpoint {
+        BN254.G1Point previousPubkeyG1;
+        bytes32 previousPubkeyHash;
+        uint32 blockNumber;
```
ChaoticWalrus, line 45: one edge case to consider is signatures on the boundary
e.g. if the operator signs a message with one key, changes keys, and signs a message with the new key, all within a single block.
the description you have looks technically accurate, and I feel like we probably just can't perfectly solve this problem? so if you care about historical keys (e.g. for slashing) you should just treat both keys as owned by the operator during this block, right?

ChaoticWalrus (changes requested):
Looks quite good. I have a few questions + minor suggestions but this is on the right track.

src/BLSApkRegistry.sol
0x0aa0: no storage on EIP712

ChaoticWalrus: it doesn't have storage, resolving.

src/RegistryCoordinator.sol
ChaoticWalrus: discussed in-person, no sanitization seems OK, not in initializer is also acceptable to me (although not my ideal), no event is something I can live with but probably the most desirable of these

src/interfaces/IBLSApkRegistry.sol
ChaoticWalrus: talked in-person, concluded this isn't possible because the existing check prevents deregistering and re-registering in the same block (which is the only way to change your key). specific code:
`lastDeregistrationTimestamp[operator] + deregistrationCooldown < block.timestamp`
note in particular the strict inequality

ChaoticWalrus (approved):
The new view functions LGTM. They differ a bit from the pattern elsewhere due to the historical storage model being different here, but as far as I can tell it's all written correctly. Adding some unit tests that verify the behavior of these view functions would be 🔥 but I'm not going to require that for my approval on this PR.
The nits that I would have considered blocking on have been addressed.
Nice work 👍

PR #251, opened 2024-05-03
fix: add init function for staleStakesForbidden in BLSSignatureChecker

stevennevins:


src/BLSSignatureChecker.sol
```diff
@@ -15,13 +17,13 @@ import {BN254} from "./libraries/BN254.sol";
  * @notice Terms of Service: https://docs.eigenlayer.xyz/overview/terms-of-service
  * @notice This is the contract for checking the validity of aggregate operator signatures.
  */
-contract BLSSignatureChecker is IBLSSignatureChecker {
+contract BLSSignatureChecker is Initializable, IBLSSignatureChecker {
```
ChaoticWalrus: I am somewhat worried about the storage layout change here.
Can we do without this import? I think the only way this storage layout is backwards-compatible is if you're upgrading a contract that inherits from Initializable directly before inheriting from this contract? Seems like a large potential footgun to get an initializer for a single boolean

ChaoticWalrus (changes requested):
The idea is good to me, but I think the storage layout change is likely worth blocking this on.

src/BLSSignatureChecker.sol
stevennevins: makes sense, anything that is also initializable shouldn't have shift, but i understand the concern

ChaoticWalrus: the linearization of the inheritance is ultimately what matters, so just inheriting Initializable from elsewhere isn't a guarantee either 😞

0x0aa0 (approved):
Am much happier without the initializer here as it requires no change to DASM. Seems like we just need to communicate to AVS teams that this is something they need to flip the switch on some time after running AVSSync

ChaoticWalrus (approved):
Works for me now.
Thanks for accommodating my request here.

src/BLSSignatureChecker.sol
stevennevins: yeah that's true, seems like it could be a foot gun

PR #252, opened 2024-05-07
feat: ecdsa key rotation

stevennevins:


src/unaudited/ECDSAStakeRegistry.sol
```diff
@@ -127,6 +140,37 @@ contract ECDSAStakeRegistry is
...
+        return address(uint160(_operatorSigningKeyHistory[_operator].latest()));
+    }
+
+    /**
+     * @notice Retrieves the latest signing key for a given operator at a specific block number.
+     * @param _operator The address of the operator.
+     * @param _blockNumber The block number to get the operator's signing key.
+     * @return The signing key of the operator at the given block.
+     */
+    function getLastestOperatorSigningKeyAtBlock(
```
ChaoticWalrus: how is it both 'latest' and 'at a specific block number'?
is it possible for an operator to have multiple keys in the same block?
how does this handle any 'edge case' behavior of an operator changing keys in the middle of a block?

ChaoticWalrus: to answer my own second question, the storage will erase any previous keys at the same block, since the `Checkpoints.History.push()` operation is being used any time a key update happens
(see code here https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/6b9807b0639e1dd75e07fa062e9432eb3f35dd8c/contracts/utils/CheckpointsUpgradeable.sol)

for the BLS contracts, we introduced a minimum period between key updates, which prevents multiple key updates in the same block, so there is no possibility (albeit pathological) of a user

- updating a key
- signing with the new key
- and updating a key again

all within the same block, as the storage model would have no evidence of this key.

Either way, it seems that we need to assume both keys are active on the boundary block.

src/unaudited/ECDSAStakeRegistryStorage.sol
```diff
@@ -27,14 +29,19 @@ abstract contract ECDSAStakeRegistryStorage is ECDSAStakeRegistryEventsAndErrors
     /// @notice Defines the duration after which the stake's weight expires.
     uint256 internal _stakeExpiry;
 
+    /// @notice Maps an operator to their signing key history using checkpoints
+    mapping(address => CheckpointsUpgradeable.History)
+        internal _operatorSigningKeyHistory;
+
```
ChaoticWalrus, line 35: do you care at all about storage compatibility for upgrades? or not concerned yet because we can simply redeploy instead of upgrading

ChaoticWalrus (changes requested):
There's one edge case that -- while it doesn't seem likely to be exploited -- could theoretically let a deserving party avoid slashing, by changing keys twice in the same block and signing with the "ephemeral" key, which would never make it into historical storage.
I believe it could be solved by preventing multiple signing key updates within the same block. Implementing this check within the `_updateOperatorSigningKey` function seems most appropriate to me, personally.

Otherwise the PR LGTM.

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: Nice call on the function name, that was a typo

I think it's okay not to have both be valid in the same block

The behavior is that the operator's last key registered in a block would be the valid key for the operator, which seems sufficient to me.  I thought about adding the delay, but it didn't seem necessary

src/unaudited/ECDSAStakeRegistryStorage.sol
stevennevins: Yeah not concerned about it rn, once we have people using it we can be cautious about these changes

ChaoticWalrus: cool, works for me. I'll resolve this thread then.

src/unaudited/ECDSAStakeRegistry.sol
ChaoticWalrus: I think I'm implicitly seeing the storage as relevant for slashing operators.
if that's not needed, then this all seems fine, but if the intention is to use this storage in fraudproofs (especially for slashing) then the history needs to be complete, and _for the purposes of fraudproofs_, it seems that we have to assume that the operator had both keys active on the boundary.
I agree that practically, the operator can (and should) only be able to sign with one key at any time.

stevennevins: Here the binary search logic will only return the most recent update when multiple exist in a single block.  In effect it's like the first one didn't happen from the perspective of verifying a message from the operator

stevennevins: We can prevent doing two updates in a block though if it's still a concern

ChaoticWalrus: so I think there's two aspects to this:

1. there's a theoretical concern about multiple updates in the same block
2. in fraudproof usage, the binary search logic cannot be relied upon to give the correct answer for an operator's key, because fraudproofs must consider the old key to be active on the boundary

I realize (1) is absolutely a pathological kind of bug but I'd rather not have it present.
For (2), nothing is technically wrong, but users will need clear guidance on this. I believe the same kind of issue is present in the BLS contracts?
Although now that I consider it, the BLS contracts disallow using the current block as the reference block number, which I think actually addresses both of these issues? Since this makes the concerning aspect of (1) -- signing with a key that was only "active" for part of a block -- impossible, and means the operator's signing key is definitive for any block before it is used, which solves (2)

stevennevins: Yeah, that would be another approach. I can add the check to _checkSignatures that the reference block is older than the current block

stevennevins: Addressed signing at the current block height [f6942c7](https://github.com/Layr-Labs/eigenlayer-middleware/pull/252/commits/f6942c73880c3ce5441ff47b4abc3561d2a70722)

stevennevins:
@ChaoticWalrus addressed the PR comments and added two more unit tests for reference block number checks

src/unaudited/ECDSAStakeRegistry.sol
stevennevins: fixed typo in function signature [8854cf9](https://github.com/Layr-Labs/eigenlayer-middleware/pull/252/commits/8854cf983444167b64d9e7a484c5a532e9e2b88a)

ChaoticWalrus: LGTM, thanks

ChaoticWalrus (approved):
Looks like everything I flagged before has been addressed. Thanks! Giving this approval now.

PR #256, opened 2024-05-08
test: fix flaky fuzz tests hitting the default operator

stevennevins:
Adding the the filter fuzz inputs like the happy path test

We were sometimes trying to re-register the default operator and this cause the revert tests to revert for the incorrect reason occasionally (or at high fuzz runs)


ChaoticWalrus (approved):
LGTM, seems appropriate, ty

PR #266, opened 2024-06-06
perf: refactor modifiers to use internal functions for stake registry

stevennevins:


wadealexc (approved):
Looks great! I'm seeing this saves ~1kB bytecode in the StakeRegistry

PR #274, opened 2024-06-16
feat: reward initiator for ECDSAServiceManagerBase

Gajesh2007:


src/unaudited/ECDSAServiceManagerBase.sol
```diff
@@ -234,6 +249,20 @@ abstract contract ECDSAServiceManagerBase is
...
+    }
+
+    function _setRewardsInitiator(address newRewardsInitiator) internal {
+        emit RewardsInitiatorUpdated(rewardsInitiator, newRewardsInitiator);
+        rewardsInitiator = newRewardsInitiator;
+    }
+
     // storage gap for upgradeability
     // slither-disable-next-line shadowing-state
     uint256[50] private __GAP;
```
stevennevins: Need to update the storage gap for the new state

src/unaudited/ECDSAServiceManagerBase.sol
```diff
@@ -42,6 +41,17 @@ abstract contract ECDSAServiceManagerBase is
         _;
     }
 
+    /**
+     * @dev Ensures that the function is only callable by the `rewardsInitiator`.
+     */
+    modifier onlyRewardsInitiator() {
+        require(
```
stevennevins, line 53: Can you put this in an internal function

```
modifier onlyRewardInitiator() {
    _checkRewardsInitiator();
    _;
}
```

src/unaudited/ECDSAServiceManagerBase.sol
```diff
@@ -168,9 +181,11 @@ abstract contract ECDSAServiceManagerBase is
                 address(this),
                 rewardsSubmissions[i].amount
             );
+            uint256 allowance =
+                rewardsSubmissions[i].token.allowance(address(this), rewardsCoordinator);
             rewardsSubmissions[i].token.approve(
                 rewardsCoordinator,
-                rewardsSubmissions[i].amount
+                rewardsSubmissions[i].amount + allowance
```
stevennevins, line 192: nice thanks for doing this as well

src/unaudited/ECDSAServiceManagerBase.sol
Gajesh2007: by how much?

src/unaudited/ECDSAServiceManagerBase.sol
Gajesh2007: it's just one require() check. do u think we need to put it in another func?

src/unaudited/ECDSAServiceManagerBase.sol
stevennevins: Needs to be 49 now.  

Basically you allocate up front 50 slots so when you make storage changes you don't shift the storage of contracts in the inheritance graph. Then for each storage slot you use in the contract you deduct 1 from your gap

src/unaudited/ECDSAServiceManagerBase.sol
stevennevins: Yeah that way if we go to add the modifier in the future we don't forgot to make this change to reduce the duplication of the inlined code

Gajesh2007:
ready for merge @stevennevins

stevennevins:
lgtm ty

stevennevins (approved):

PR #275, opened 2024-06-20
fix: service manager storage layout

0x0aa0:
Updated inheritance such that the EigenDAServiceManager maintains its current storage layout. This broke in the separation of the ServiceManagerBase from its ServiceManagerBaseStorage


8sunyuan:
imageimage
Checked storage and LGTM, the other commit was taken from before the ServiceManagerBaseStorage contract was created

8sunyuan (approved):

PR #279, opened 2024-06-26
feat: operator-sets-3-registration

stevennevins:


src/ServiceManagerBase.sol
```diff
@@ -122,16 +129,42 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
      * @notice Forwards a call to EigenLayer's AVSDirectory contract to confirm operator deregistration from the AVS
      * @param operator The address of the operator to deregister.
      */
-    function deregisterOperatorFromAVS(address operator) public virtual onlyRegistryCoordinator {
+    function deregisterOperatorFromAVS(
```
ypatil12, line 132: Should we still implement this function in our canonical example?

src/ServiceManagerBase.sol
```diff
@@ -146,7 +179,11 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
      * @dev No guarantee is made on uniqueness of each element in the returned array.
      *      The off-chain service should do that validation separately
      */
-    function getRestakeableStrategies() external view returns (address[] memory) {
+    function getRestakeableStrategies()
```
ypatil12, line 182: We don't need this function anymore for operator set based AVSs, since this is provided via the MetadataURI

src/ServiceManagerBase.sol
```diff
@@ -178,34 +218,41 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
      * @dev No guarantee is made on whether the operator has shares for a strategy in a quorum or uniqueness
      *      of each element in the returned array. The off-chain service should do that validation separately
      */
-    function getOperatorRestakedStrategies(address operator)
-        external
-        view
-        returns (address[] memory)
-    {
+    function getOperatorRestakedStrategies(
```
ypatil12, line 221: Not needed for operator set release, since computed via operator<>operatorSet indexing along with metadataURI

ypatil12 (commented):
Minor comments

src/ServiceManagerBase.sol
stevennevins: gotcha. Were we envisioning that we would only have operator sets going forward, or support two ServiceManager contracts?

stevennevins:
closed in favor of https://github.com/Layr-Labs/eigenlayer-middleware/pull/301

PR #280, opened 2024-06-28
fix: correct index get operator restakable strategies

stevennevins:


ChaoticWalrus (approved):
LGTM.
I'm assuming the same issue isn't present in the BLS contracts?

stevennevins:
> LGTM. I'm assuming the same issue isn't present in the BLS contracts?

Yeah it's not, this was a mistake when I was replicating the functionality

PR #285, opened 2024-07-17
Update LICENSE

gpsanant:


ChaoticWalrus (approved):
As I understand Git the proper thing to do here would be to cherry-pick the relevant commit to this branch. I think this is OK though; the content exactly matches what's on the `dev` branch at least.

stevennevins (approved):

PR #286, opened 2024-07-22
feat: operator set migration-1-migration

stevennevins:


src/ServiceManagerBase.sol
```diff
@@ -135,6 +136,145 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
         _setRewardsInitiator(newRewardsInitiator);
     }
 
+    function migrateToOperatorSets() external onlyOwner {
```
ypatil12: How gassy is this function? What is the maximum number of operators & operator sets that can be migrated? If there is a limit, should we be specifying a list of operators to migrate?

src/ServiceManagerBase.sol
```diff
@@ -135,6 +136,145 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+        _migrateToOperatorSets();
+    }
+
+    /**
+     * @notice Merges two sorted arrays using the merge sort algorithm
+     * @param left The first sorted array
+     * @param right The second sorted array
+     * @return The merged sorted array
+     */
+    function mergeSortedArrays(uint256[] memory left, uint256[] memory right) internal pure returns (uint256[] memory) {
```
ypatil12: Nit: _ for internal function

src/ServiceManagerBase.sol
```diff
@@ -135,6 +136,145 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+                    filteredOperators[count++] = allOperators[i];
+                }
+            }
+            // Resize array to remove empty slots
+            assembly { mstore(filteredOperators, count) }
+            allOperators = filteredOperators;
+
+            operatorSetIds = new uint32[][](allOperators.length);
+            // Loop through each unique operator to get the quorums they are registered for
+            for (uint256 i = 0; i < allOperators.length; i++) {
```
ypatil12: This step can occur outside the step1 loop? Since we can:

For each quorum
- Get all operators
- Insert into all operators array

After getting operators across all quorums:

For each operator in all operators
- Get quorum bitmap array

ypatil12 (commented):
Instead of doing mergesort, should we just add a storage variable(s) that go from operatorIndex to operator (and back) so then we can key into each operator (based on total number of operators) and then add

Still think there may be questions around gas & if some AVSs would want to specify a list of operators or batch it over multiple tx's

src/ServiceManagerBase.sol
stevennevins: [97fb672](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/97fb67286247415c3dd9344f024e0a86283a6562)

src/ServiceManagerBase.sol
stevennevins: Yeah i think for EigenDA it could be an issue atm.  I ran some gas tests for 4 quorums and ~100 operators per quorum and it was on the order of 100s million gas.  I think the function could be optimized though since i haven't really focused on that yet.

I refactored to a view function like you suggested but left it being called internally to the `migrate()` for now to see if we can optimize to get to something reasonable.  If we can't get it down to around 10 million gas then we should do multiple txs

src/ServiceManagerBase.sol
stevennevins: Yeah good call!

src/ServiceManagerBase.sol
stevennevins: I did a bit more testing this morning to split out the view and pass in the state that needed to be updated vs. doing the view calls in the state-changing function, and it's ~21 million gas based on 3 quorums and 200 operators per quorum. I consider this worst case scenario if EigenDA has reAlt go live on mainnet and they have about the number of operators in the Eigen and Eth quorums

src/ServiceManagerBase.sol
stevennevins: [8b46f83](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/8b46f83ea9e608abfe933a40e09f0ed48c6a8933)

src/ServiceManagerBase.sol
stevennevins: Migrating can now be broken up over multiple transactions if the situation arises to prevent potential issues with gas costs exceed the block size
[3fcc68e](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/3fcc68e1370cf86c893790f0e44d62e7ad714e7d)

src/ServiceManagerBase.sol
```diff
@@ -135,6 +137,114 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
         _setRewardsInitiator(newRewardsInitiator);
     }
 
+    function migrateOperatorSetIds(uint32[] memory operatorSetsToCreate) external onlyOwner{
```
ypatil12: Nit: `migrateAndCreateOperatorSetIds` ? Function does two things -> becomes an operator set avs and creates operator set Ids

src/ServiceManagerBase.sol
```diff
@@ -135,6 +137,114 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+        _migrateCreateOperatorSetIds(operatorSetsToCreate);
+    }
+
+    function migrateToOperatorSets(uint32[][] memory operatorSetIds, address[] memory operators) external onlyOwner {
+        require(!migrationFinalized, "SerivceManager: Migration Already Finalized");
```
ypatil12, line 146: Is this require statement needed? Seems like it's not giving much of a guarantee?

ypatil12: Or, do we want to just to be explicit when avs has migrated ig?

src/ServiceManagerBase.sol
```diff
@@ -135,6 +137,114 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+            uint32[] memory quorums = new uint32[](quorumBytesArray.length);
+            for (uint256 j = 0; j < quorumBytesArray.length; j++) {
+                quorums[j] = uint32(uint8(quorumBytesArray[j]));
+            }
+            operatorSetIds[i] = quorums;
+        }
+
+    }
+
+    function _mergeSortedArrays(address[] memory left, address[] memory right) internal pure returns (address[] memory) {
```
ypatil12: Nit: potentially can move this to a library?

src/ServiceManagerBase.sol
```diff
@@ -135,6 +137,114 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+    function _migrateCreateOperatorSetIds(uint32[] memory operatorSetIdsToCreate) internal {
+        _avsDirectory.becomeOperatorSetAVS();
+        AVSDirectory(address(_avsDirectory)).createOperatorSets(operatorSetIdsToCreate);
+    }
+
+    function _migrateToOperatorSets(uint32[][] memory operatorSetIds, address[] memory operators) internal {
+        AVSDirectory(address(_avsDirectory)).migrateOperatorsToOperatorSets(operators, operatorSetIds);
+    }
+
+    function getOperatorsToMigrate() public view returns (uint32[] memory operatorSetIdsToCreate, uint32[][] memory operatorSetIds, address[] memory allOperators) {
```
ypatil12, line 164: Can we add fuzz tests for this function?

ypatil12 (commented):
Minor comments

src/ServiceManagerBase.sol
stevennevins: Before it wasn't possible to do two transactions because `becomeOperatorSetAVS` blocked you from doing the second call and effectively resulted in a finalization guarantee of the migration.  But now that we needed to do 2 calls, the becomeOperatorSetAVS check happens separately in the `migrateOperatorSetIds` function above

src/ServiceManagerBase.sol
stevennevins: yeah that's reasonable

src/ServiceManagerBase.sol
stevennevins: [39d56b1](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/39d56b1996e38ea7d6437b9c83e566909c5b1d23)

src/ServiceManagerBase.sol
stevennevins: [2ee11df](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/2ee11dfc9cc92429d9b0b7d75b383e1fb8f7631a)

src/ServiceManagerBase.sol
stevennevins: Added a fuzz test for registering random sets of operators and viewing the migration data. checked the invariants of the _registerRandomOperators [6f37c47](https://github.com/Layr-Labs/eigenlayer-middleware/pull/286/commits/6f37c4710a9e441a386f0334db4924199611f696)

ypatil12 (approved):

PR #287, opened 2024-07-25
feat: operator set migration-2-create quorum

stevennevins:


src/RegistryCoordinator.sol
```diff
@@ -732,6 +733,14 @@ contract RegistryCoordinator is
         stakeRegistry.initializeQuorum(quorumNumber, minimumStake, strategyParams);
         indexRegistry.initializeQuorum(quorumNumber);
         blsApkRegistry.initializeQuorum(quorumNumber);
+        // Check if the AVS has migrated to operator sets
+        AVSDirectory avsDirectory = AVSDirectory(serviceManager.avsDirectory());
+        if (avsDirectory.isOperatorSetAVS(address(serviceManager))){
```
ypatil12, line 738: Assume we do this check because we don't want the migration function to fail?

ypatil12 (commented):
One comment

src/RegistryCoordinator.sol
stevennevins: I think we could createOperatorSets before migration before, so I wanted to prevent that at the time

src/RegistryCoordinator.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.12;
 
 import {IPauserRegistry} from "eigenlayer-contracts/src/contracts/interfaces/IPauserRegistry.sol";
 import {ISignatureUtils} from "eigenlayer-contracts/src/contracts/interfaces/ISignatureUtils.sol";
+import {AVSDirectory} from "eigenlayer-contracts/src/contracts/core/AVSDirectory.sol";
```
ypatil12, line 6: Nit: can we import interface? Does the interface not have all functions we need?

ypatil12 (approved):
LGTM, added one minor comment, but approving

src/RegistryCoordinator.sol
stevennevins: the interface was missing a function initially.

I'll double check if it's there now

stevennevins: It was there now [b69d253](https://github.com/Layr-Labs/eigenlayer-middleware/pull/287/commits/b69d253bdf9bff683fdd3e21410b0cb9f6ca7f14)

PR #288, opened 2024-07-25
feat: operator set release

stevennevins:


src/ServiceManagerBase.sol
```diff
@@ -135,6 +140,77 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
         _setRewardsInitiator(newRewardsInitiator);
     }
 
+    function migrateAndCreateOperatorSetIds(uint32[] memory operatorSetsToCreate) external onlyOwner{
```
gpsanant: comments on all funcs

src/ServiceManagerBase.sol
```diff
@@ -135,6 +140,77 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
+        migrationFinalized = true;
+    }
+
+    function _migrateAndCreateOperatorSetIds(uint32[] memory operatorSetIdsToCreate) internal {
+        _avsDirectory.becomeOperatorSetAVS();
+        IAVSDirectory(address(_avsDirectory)).createOperatorSets(operatorSetIdsToCreate);
+    }
+
+    function _migrateToOperatorSets(uint32[][] memory operatorSetIds, address[] memory operators) internal {
+        IAVSDirectory(address(_avsDirectory)).migrateOperatorsToOperatorSets(operators, operatorSetIds);
```
gpsanant: are we not checking that the provided operators are correct?

gpsanant: there's a race condition here where operators deregister for a certain quorum inbetween when the view function is called and migrate is called

ypatil12: We can and just skip. May be cleaner

ypatil12: Regarding race condition, we can also prevent m2 deregistrations in AVSDirectory

stevennevins: yeah we should block m2 functions in core  
`if (isOperatorSetAVS[avs]==true)`

src/ServiceManagerBase.sol
stevennevins: [12eda5d](https://github.com/Layr-Labs/eigenlayer-middleware/pull/288/commits/12eda5dd36ba793b03136516d2342554b173151a)

src/ServiceManagerBase.sol
stevennevins: added additional checks [d64f4af](https://github.com/Layr-Labs/eigenlayer-middleware/pull/288/commits/d64f4af465201d66e2c581e42f28b32764a82381)

ypatil12: added pr here to block m2 dereg: https://github.com/Layr-Labs/eigenlayer-contracts/pull/688

stevennevins:
Resolved the code size issue by wrapping the library from core

PR #300, opened 2024-08-19
feat(op sets): update stakes when forceUnregister

stevennevins:


src/StakeRegistry.sol
```diff
@@ -148,6 +150,12 @@ contract StakeRegistry is StakeRegistryStorage {
     ) external onlyRegistryCoordinator returns (uint192) {
         uint192 quorumsToRemove;
 
+         address avs = address(
```
ypatil12: Could make this immutable & refer to it as serviceManager?

ypatil12: similar to what we do in registry coordinator

ypatil12 (commented):
One comment

src/StakeRegistry.sol
stevennevins: yeah good call

stevennevins: all set [9b96076](https://github.com/Layr-Labs/eigenlayer-middleware/pull/300/commits/9b96076c9460c37ba7bdab4a7be32042890380e8)

ypatil12 (approved):

PR #301, opened 2024-08-19
feat(op sets): register and deregister

stevennevins:


src/RegistryCoordinator.sol
```diff
@@ -515,7 +515,18 @@ contract RegistryCoordinator is
                 OperatorInfo({operatorId: operatorId, status: OperatorStatus.REGISTERED});
 
             // Register the operator with the EigenLayer core contracts via this AVS's ServiceManager
-            serviceManager.registerOperatorToAVS(operator, operatorSignature);
+            bool operatorSetAVS = IAVSDirectory(serviceManager.avsDirectory()).isOperatorSetAVS(address(serviceManager)); /// TODO: call avsdirectory
```
ypatil12: Stale TODO?

src/RegistryCoordinator.sol
```diff
@@ -515,7 +515,18 @@ contract RegistryCoordinator is
...
+            if (operatorSetAVS){
+                bytes memory quorumBytes = BitmapUtils.bitmapToBytesArray(quorumsToAdd);
+                uint32[] memory operatorSetIds = new uint32[](quorumBytes.length);
+                for (uint256 i = 0; i < quorumBytes.length; i++) {
+                    operatorSetIds[i] = uint8(quorumBytes[i]);
```
ypatil12, line 525: Why `uint8(quorumBytes[I])`?

src/RegistryCoordinator.sol
```diff
@@ -647,19 +658,31 @@ contract RegistryCoordinator is
...
+            }
+            serviceManager.deregisterOperatorFromOperatorSets(operator, operatorSetIds);
+
+        } else {
+            // If the operator is no longer registered for any quorums, update their status and deregister
+            // them from the AVS via the EigenLayer core contracts
+            if (newBitmap.isEmpty()) {
+                operatorInfo.status = OperatorStatus.DEREGISTERED;
+                serviceManager.deregisterOperatorFromAVS(operator);
+                emit OperatorDeregistered(operator, operatorId);
```
ypatil12, line 709: We still want to emit this event even in operatorSet-AVS case?

src/RegistryCoordinator.sol
```diff
@@ -647,19 +658,31 @@ contract RegistryCoordinator is
...
         _updateOperatorBitmap({operatorId: operatorId, newBitmap: newBitmap});
 
-        // If the operator is no longer registered for any quorums, update their status and deregister
-        // them from the AVS via the EigenLayer core contracts
-        if (newBitmap.isEmpty()) {
-            operatorInfo.status = OperatorStatus.DEREGISTERED;
-            serviceManager.deregisterOperatorFromAVS(operator);
-            emit OperatorDeregistered(operator, operatorId);
+
+        bool operatorSetAVS = IAVSDirectory(serviceManager.avsDirectory()).isOperatorSetAVS(address(serviceManager)); /// TODO: call avsdirectory
```
ypatil12: nit: stale TODO?

src/libraries/BN254.sol
```diff
@@ -62,7 +62,7 @@ library BN254 {
     ///      this is because of the (unknown to us) convention used in the bn254 pairing precompile contract
     ///      "Elements a * i + b of F_p^2 are encoded as two elements of F_p, (a, b)."
     ///      https://github.com/ethereum/EIPs/blob/master/EIPS/eip-197.md#encoding
-    function generatorG2() internal pure returns (G2Point memory) {
+    function generatorG2() external pure returns (G2Point memory) {
```
ypatil12, line 65: Is making this library external only way to decrease code size?

src/interfaces/IRegistryCoordinator.sol
```diff
@@ -137,7 +137,7 @@ interface IRegistryCoordinator {
     function registries(uint256) external view returns (address);
 
     /// @notice Returns the number of registries
-    function numRegistries() external view returns (uint256);
+    // function numRegistries() external view returns (uint256);
```
ypatil12: Why commented out?

src/libraries/BN254.sol
stevennevins: Yeah we are at ~30 bytes remaining

src/interfaces/IRegistryCoordinator.sol
stevennevins: was trying to reduce bytecode size forgot to remove the comment in interface after i added the function back

src/RegistryCoordinator.sol
stevennevins: [0285a51](https://github.com/Layr-Labs/eigenlayer-middleware/pull/301/commits/0285a513d0ec6ac1f27faec50d70eec43b1acb47)

src/RegistryCoordinator.sol
stevennevins: [0285a51](https://github.com/Layr-Labs/eigenlayer-middleware/pull/301/commits/0285a513d0ec6ac1f27faec50d70eec43b1acb47)

src/RegistryCoordinator.sol
stevennevins: i think so. I was wondering if there were any offchain dependencies on this event. it seemed reasonable to keep. This would be if they were removed deregistered from all operators sets they were a part of

src/libraries/BN254.sol
stevennevins: I did this on the functions that aren't called multiple times in loops, so the gas impact was minimal

src/interfaces/IRegistryCoordinator.sol
stevennevins: added back [cb98fc2](https://github.com/Layr-Labs/eigenlayer-middleware/pull/301/commits/cb98fc21269a95826443b6136f0096c03fbe6839)

src/RegistryCoordinator.sol
stevennevins: We can't go directly from bytes1 -> uint32

the uint8 to uint32 happens implicitly when passed to the function call

src/RegistryCoordinator.sol
ypatil12: I assume it's AVS specific. Does EigenDA use this event??

My main comment is the `OperatorDeregistered` event is not emitted at all if the AVS is an operatorSetAVS, which seems off

stevennevins: I see what you mean. Moved the OperatorRegistered event for M2 to only be emitted if they're still an M2 AVS [6a77df8](https://github.com/Layr-Labs/eigenlayer-middleware/pull/301/commits/6a77df8cf257308f4df87bde96d692ede2eff3ce)

stevennevins:
@ypatil12 when you have a sec can you take another pass here

ypatil12 (approved):

PR #307, opened 2024-09-26
onchain socket

0x0aa0:
- move socket storage to new registry
- leave event in RegCoord
- trim bytecode from error strings


src/RegistryCoordinator.sol
```diff
@@ -502,6 +501,9 @@ contract RegistryCoordinator is
             // Register the operator with the EigenLayer core contracts via this AVS's ServiceManager
             serviceManager.registerOperatorToAVS(operator, operatorSignature);
 
+            socketRegistry.setOperatorSocket(operatorId, socket);
```
gpsanant: call same internal function as updateSocket

src/SocketRegistry.sol
```diff
@@ -0,0 +1,52 @@
...
+        registryCoordinator = address(_registryCoordinator);
+    }
+
+    /// @notice sets the socket for an operator only callable by the RegistryCoordinator
+    function setOperatorSocket(bytes32 _operatorId, string memory _socket) external onlyRegistryCoordinator {
+        operatorIdToSocket[_operatorId] = _socket;
+    }
+
+    /// @notice migrates the sockets for a list of operators only callable by the owner of the RegistryCoordinator
+    function migrateOperatorSockets(bytes32[] memory _operatorIds, string[] memory _sockets) external onlyCoordinatorOwner {
```
stevennevins, line 41: Is it okay if this won't emit an event in the registry coord?

0x0aa0: yes the events are going to eventually be deprecated

stevennevins: We should keep the events around so we know when they change?

0x0aa0: Ya good point we can leave those but we wont need it in the migration function

stevennevins (approved):

PR #309, opened 2024-10-12
Reward Submission Example

0x0aa0:
an example of a rewards submission


afkbyte:
Thanks so much arch!!

PR #313, opened 2024-10-23
feat: ejection policy change

0x0aa0:
Changes the EjectionManager behavior to eject the operator that trips the ratelimit but no further

### DIFF
--- a/src/EjectionManager.sol
+++ b/src/EjectionManager.sol
@@ -28,6 +28,9 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
     /// @notice Ratelimit parameters for each quorum
     mapping(uint8 => QuorumEjectionParams) public quorumEjectionParams;
 
+    /// @notice mapping from quorum number to operator stake cap percentage
+    mapping(uint8 => uint256) public operatorStakeCapPercent;
+
     constructor(
         IRegistryCoordinator _registryCoordinator,
         IStakeRegistry _stakeRegistry
@@ -46,7 +49,8 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
     function initialize(
         address _owner,
         address[] memory _ejectors,
-        QuorumEjectionParams[] memory _quorumEjectionParams
+        QuorumEjectionParams[] memory _quorumEjectionParams,
+        uint256[] memory _operatorStakeCapPercent
     ) external initializer {
         _transferOwnership(_owner);
         for(uint8 i = 0; i < _ejectors.length; i++) {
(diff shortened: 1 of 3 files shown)


src/EjectionManager.sol
```diff
@@ -144,17 +171,25 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
...
+        operatorStakeCapPercent[_quorumNumber] = _operatorStakeCapPercent;
+        emit OperatorStakeCapPercentSet(_quorumNumber, _operatorStakeCapPercent);
+    }
+
     /**
      * @notice Returns the amount of stake that can be ejected for a quorum at the current block.timestamp
      * @param _quorumNumber The quorum number to view ejectable stake for
      */
-    function amountEjectableForQuorum(uint8 _quorumNumber) public view returns (uint256) {
+    function amountEjectableForQuorum(uint8 _quorumNumber) public view returns (uint256 ejectableStake, uint256 totalStake) {
```
stevennevins: It feels like it would be easier if we:
prevented them from registering / updating their stake to an amount greater than the stake cap in the first place instead of ejecting an amount of stake from the operator(s) after the fact

stevennevins: Discussed offline. Variable names were confusing

src/EjectionManager.sol
```diff
@@ -83,11 +83,18 @@ contract EjectionManager is IEjectionManager, OwnableUpgradeable{
                     quorumEjectionParams[quorumNumber].rateLimitWindow > 0 &&
                     stakeForEjection + operatorStake > amountEjectable
                 ){
-                    stakeEjectedForQuorum[quorumNumber].push(StakeEjection({
-                        timestamp: block.timestamp,
-                        stakeEjected: stakeForEjection
-                    }));
                     ratelimitHit = true;
+
```
stevennevins, line 87: In the above, if statement, it feels like we could simplify the logic flow by inverting the condition and pulling out the `rateLimitWindow` check to a guard clause

`

            uint32 numEjectedOperators;

            bool ratelimitHit;
            for(uint8 j = 0; j < _operatorIds[i].length; ++j) {
                uint256 operatorStake = stakeRegistry.getCurrentStake(_operatorIds[i][j], quorumNumber);
                if(quorumEjectionParams[quorumNumber].rateLimitWindow == 0) {
                    break;
                }
                if (isEjector[msg.sender] && !ratelimitHit) {
                    stakeForEjection += operatorStake;
                    ++numEjectedOperators;
                    registryCoordinator.ejectOperator(
                        registryCoordinator.getOperatorFromId(_operatorIds[i][j]),
                        abi.encodePacked(quorumNumber)
                    );

                    emit OperatorEjected(_operatorIds[i][j], quorumNumber);

                    if (stakeForEjection > amountEjectable) {
                        stakeEjectedForQuorum[quorumNumber].push(StakeEjection({
                            timestamp: block.timestamp,
                            stakeEjected: stakeForEjection
                        }));
                        break;
                    }
                }
            }

            emit QuorumEjection(numEjectedOperators, ratelimitHit);
        }
`

stevennevins (approved):

PR #317, opened 2024-11-13
feat: track total slashable stake and total delegated stake per quorum

stevennevins:


src/RegistryCoordinator.sol
```diff
@@ -824,7 +839,14 @@ contract RegistryCoordinator is
...
+        // Initialize stake registry based on stake type
+        if (stakeType == IStakeRegistry.StakeType.TOTAL_DELEGATED) {
+            stakeRegistry.initializeDelegatedStakeQuorum(quorumNumber, minimumStake, strategyParams);
+        } else if (stakeType == IStakeRegistry.StakeType.TOTAL_SLASHABLE) {
+            stakeRegistry.initializeSlashableStakeQuorum(quorumNumber, minimumStake, lookAheadPeriod, strategyParams);
+        }
+
         indexRegistry.initializeQuorum(quorumNumber);
         blsApkRegistry.initializeQuorum(quorumNumber);
         // Check if the AVS has migrated to operator sets
```
ypatil12, line 852: Need to update submodule to latest slashing branch. We're still checking if AVS is an opSet AVS in AVSD

src/StakeRegistry.sol
```diff
@@ -493,6 +520,23 @@ contract StakeRegistry is StakeRegistryStorage {
...
     }
 
+    /// Returns total Slashable stake for an operator per strategy that can have the weights applied based on strategy multipliers
+    function _getSlashableStakePerStrategy(uint8 quorumNumber, address operator) internal view returns (uint256[] memory) {
+        address[] memory operators = new address[](1);
+        operators[0] = operator;
+        uint32 beforeTimestamp = uint32(block.timestamp + slashableStakeLookAheadPerQuorum[quorumNumber]);
+
+        (,uint256[][] memory slashableShares) = IAllocationManager(serviceManager.allocationManager())
+            .getMinDelegatedAndSlashableOperatorShares(
```
ypatil12, line 530: Can use new method? Otherwise makes sense

src/StakeRegistry.sol
```diff
@@ -503,18 +547,27 @@ contract StakeRegistry is StakeRegistryStorage {
...
-        for (uint256 i = 0; i < stratsLength; i++) {
-            // accessing i^th StrategyParams struct for the quorumNumber
-            strategyAndMultiplier = strategyParams[quorumNumber][i];
-
-            // add the weight from the shares for this strategy to the total weight
-            if (strategyShares[i][0] > 0) {
-                weight += uint96(strategyShares[i][0] * strategyAndMultiplier.multiplier / WEIGHTING_DIVISOR);
+        uint256[] memory strategyShares;
+
+        if (stakeTypePerQuorum[quorumNumber]== StakeType.TOTAL_SLASHABLE) {
```
ypatil12, line 552: Are we also going to have ONLY slashing compatible contracts? IMO we should.

ypatil12: I assume rn we're in this weird stage of supporting both paths for AVSs like EigenDA to upgrade to. But after upgrading we can just deprecate these hybrid M2/Slashing AVS contracts

ypatil12 (commented):
First pass - looking clean so far. Can we update the submodule base?

src/interfaces/IStakeRegistry.sol
```diff
@@ -14,10 +14,9 @@ interface IStakeRegistry is IRegistry {
 
     enum StakeType {
```
gpsanant: can we define an enum outside of an interface or nah?

src/StakeRegistry.sol
```diff
@@ -207,14 +207,36 @@ contract StakeRegistry is StakeRegistryStorage {
...
+        _totalStakeHistory[quorumNumber].push(StakeUpdate({
+            updateBlockNumber: uint32(block.number),
+            nextUpdateBlockNumber: 0,
+            stake: 0
+        }));
+    }
+
+
+    /// @notice Initialize a new quorum and push its first history update
+    function initializeSlashableStakeQuorum(
```
gpsanant, line 229: hm, feel like we should combine this with the other intialize and only do slashable/delegated discrimination inside this contract, not at the interface level

src/StakeRegistryStorage.sol
```diff
@@ -52,12 +52,12 @@ abstract contract StakeRegistryStorage is IStakeRegistry {
     mapping(uint8 => StrategyParams[]) public strategyParams;
     mapping(uint8 => IStrategy[]) public strategiesPerQuorum;
 
-    StakeType public stakeType;
+    mapping(uint8 => StakeType) public stakeTypePerQuorum;
 
-    uint32 public slashableStakeLookAhead;
```
gpsanant, line 57: gap?

gpsanant: my b

src/StakeRegistry.sol
```diff
@@ -224,32 +246,37 @@ contract StakeRegistry is StakeRegistryStorage {
...
 
     /**
-     * @notice Sets the stake type for the registry
+     * @notice Sets the stake type for the registry for a specific quorum
+     * @param quorumNumber The quorum number to set the stake type for
      * @param _stakeType The type of stake to track (TOTAL_DELEGATED, TOTAL_SLASHABLE, or BOTH)
      */
-    function setStakeType(StakeType _stakeType) external onlyCoordinatorOwner {
-        _setStakeType(_stakeType);
+    function setStakeType(uint8 quorumNumber, StakeType _stakeType) external onlyCoordinatorOwner {
```
gpsanant, line 260: feels kinda unsafe setting without lookahead, not too opinionated tho. a comment would be good

src/StakeRegistry.sol
```diff
@@ -503,18 +547,27 @@ contract StakeRegistry is StakeRegistryStorage {
...
+            for (uint256 i = 0; i < stratsLength; i++) {
+                strategyAndMultiplier = strategyParams[quorumNumber][i];
+                if (strategyShares[i] > 0) {
+                    weight += uint96(strategyShares[i] * strategyAndMultiplier.multiplier / WEIGHTING_DIVISOR);
+                }
+            }
+        } else {
+            /// M2 Concept of delegated stake
+            strategyShares = delegation.getOperatorShares(operator, strategiesPerQuorum[quorumNumber]);
+            for (uint256 i = 0; i < stratsLength; i++) {
```
gpsanant: duplicated loop move outside conditional

src/RegistryCoordinator.sol
stevennevins: yup will update

src/StakeRegistry.sol
stevennevins: yeah was waiting for it to be added will update today

src/StakeRegistry.sol
stevennevins: it's not a legacy consideration pre/post-upgrade. Total delegated is a feature we want to offer in the future according to our blog posts

src/interfaces/IStakeRegistry.sol
stevennevins: yup will move

src/StakeRegistry.sol
stevennevins: based on our other comments we can align to have overloaded functions that have unused parameters depending the stake type

src/StakeRegistry.sol
stevennevins: can you provide a bit more color, it's not clear what you mean

src/StakeRegistry.sol
stevennevins: I think this is fine since it's only callable by the registry coordinator and the initialize call for slashable sets both the stake type to slashable and the lookahead

src/RegistryCoordinator.sol
stevennevins: I might prefer to do that in a follow-up PR. There were a lot of breaking changes to pull in

src/interfaces/IStakeRegistry.sol
stevennevins: [b9dccc5](https://github.com/Layr-Labs/eigenlayer-middleware/pull/317/commits/b9dccc5b0b5b441a606438a9d775b44ebb822a98)

src/StakeRegistry.sol
ypatil12: basically do:

```solidity
for (uint256 i = 0; I < stratsLength; i++) {
       If slashable, get strategyShares one way. Else Get the other way. 
       
       weight +=...
}
```

src/interfaces/IStakeRegistry.sol
```diff
@@ -6,18 +6,19 @@ import {IStrategy} from "eigenlayer-contracts/src/contracts/interfaces/IStrategy
...
+enum StakeType {
+    TOTAL_DELEGATED,
+    TOTAL_SLASHABLE
+}
+
 /**
  * @title Interface for a `Registry` that keeps track of stakes of operators for up to 256 quorums.
  * @author Layr Labs, Inc.
  */
 interface IStakeRegistry is IRegistry {
```
ypatil12, line 18: In the core our interface standard is now:

```solidity
IContractTypes {}

IContractEvents{}

IContractErrors{}

IContract is IContractTypes/Errors/Events {}
```

ypatil12: Also, can we move middleware to using custom errors? Not blocking, but we should do this for audit

src/RegistryCoordinator.sol
```diff
@@ -87,11 +87,15 @@ contract RegistryCoordinator is
         uint256 _initialPausedStatus,
         OperatorSetParam[] memory _operatorSetParams,
```
ypatil12, line 88: Nit: This has a different meaning in core. Are we OK with keeping this naming? Feels off

test/unit/RegistryCoordinatorUnit.t.sol
```diff
@@ -34,7 +34,7 @@ contract RegistryCoordinatorUnitTests is MockAVSDeployer {
 
     // Emitted when an operator pubkey is removed from a set of quorums
     event OperatorRemovedFromQuorums(
```
ypatil12, line 36: No tests for validating `quorumCount`?

src/StakeRegistry.sol
ypatil12: Are we going the unused parameters route?

ypatil12, line 260: What happens if a user wants to transition a quorum from being TDD to TSS? Is this possible (seems like no given current RegCord implementation)? Is this a valid use case?

test/unit/StakeRegistryUnit.t.sol
```diff
@@ -100,7 +100,7 @@ contract StakeRegistryUnitTests is MockAVSDeployer, IStakeRegistryEvents {
         nextQuorum++;
```
ypatil12, line 101: Can we add assertions on the storage changes for each stake type?

ypatil12: Also assertions on events

test/unit/StakeRegistryUnit.t.sol
```diff
@@ -2198,7 +2198,7 @@ contract StakeRegistryUnitTests_weightOfOperatorForQuorum is StakeRegistryUnitTe
         // create a valid quorum
         cheats.prank(address(registryCoordinator));
         uint8 quorumNumber = nextQuorum;
-        stakeRegistry.initializeQuorum(quorumNumber, 0, /* minimumStake */ strategyParams);
+        stakeRegistry.initializeDelegatedStakeQuorum(quorumNumber, 0, /* minimumStake */ strategyParams);
```
ypatil12, line 2253: We don't have tests for initializing slashable stake quorum?

ypatil12 (commented):
Testing comments

src/StakeRegistry.sol
stevennevins: I'm not a huge fan of unused params route

src/StakeRegistry.sol
stevennevins: ahh gotcha

src/interfaces/IStakeRegistry.sol
stevennevins: yeah was talking with alex about this. i was gonna do it when we hit code size issues again

src/RegistryCoordinator.sol
stevennevins: not sure, i would avoid the churn. pretty sure this existed before the term operator set in core lol

src/StakeRegistry.sol
stevennevins: Not currently supported but should be possible to switch between them if you update stakes for the quorum for everyone afterward

src/StakeRegistry.sol
stevennevins: all set 4ed4337

test/unit/RegistryCoordinatorUnit.t.sol
stevennevins: Not sure there ever were any.  Will make a note to add them in a follow up

ypatil12 (approved):

PR #323, opened 2024-11-27
docs: fix typos and add note that large array may cause revert

nadir-akhtar:
This PR addresses informational findings for the Rewards v2 internal security review. Specifically, this PR:
* Changes `callabe` to `callable` in various comments
* Adds a note that a large submissions array may result in hitting the gas limit, and to reattempt with a smaller array if so

### DIFF
--- a/src/ServiceManagerBase.sol
+++ b/src/ServiceManagerBase.sol
@@ -82,12 +82,14 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
      * @notice Creates a new rewards submission to the EigenLayer RewardsCoordinator contract, to be split amongst the
      * set of stakers delegated to operators who are registered to this `avs`
      * @param rewardsSubmissions The rewards submissions being created
-     * @dev Only callabe by the permissioned rewardsInitiator address
+     * @dev Only callable by the permissioned rewardsInitiator address
      * @dev The duration of the `rewardsSubmission` cannot exceed `MAX_REWARDS_DURATION`
      * @dev The tokens are sent to the `RewardsCoordinator` contract
      * @dev Strategies must be in ascending order of addresses to check for duplicates
      * @dev This function will revert if the `rewardsSubmission` is malformed,
      * e.g. if the `strategies` and `weights` arrays are of non-equal lengths
+     * @dev This function may fail to execute with a large number of submissions due to gas limits. Use a
+     * smaller array of submissions if necessary.
      */
     function createAVSRewardsSubmission(
         IRewardsCoordinator.RewardsSubmission[] calldata rewardsSubmissions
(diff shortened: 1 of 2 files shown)


src/ServiceManagerBase.sol
```diff
@@ -117,13 +119,15 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
      * @param operatorDirectedRewardsSubmissions The operator-directed rewards submissions being created.
-     * @dev Only callabe by the permissioned rewardsInitiator address
+     * @dev Only callable by the permissioned rewardsInitiator address
      * @dev The duration of the `rewardsSubmission` cannot exceed `MAX_REWARDS_DURATION`
      * @dev The tokens are sent to the `RewardsCoordinator` contract
      * @dev This contract needs a token approval of sum of all `operatorRewards` in the `operatorDirectedRewardsSubmissions`, before calling this function.
      * @dev Strategies must be in ascending order of addresses to check for duplicates
      * @dev Operators must be in ascending order of addresses to check for duplicates.
      * @dev This function will revert if the `operatorDirectedRewardsSubmissions` is malformed.
+     * @dev This function may fail to execute with a large number of submissions due to gas limits. Use a
```
stevennevins, line 129: what is the approximate size of a large number of submissions that would fail?

nadir-akhtar: Haven't done gas testing so I don't have that number. @0xrajath do you have an estimate? 

@stevennevins are you asking so that we can include that value in the comment?

stevennevins: if the number is sufficiently large that it's more theoretical than a practical consideration, I'm okay without being explicit in the comment. If it's on the order of 10s of reward submissions, then i think we should document it

nadir-akhtar: I can't say how many submissions are expected. That said, I recommend including the comment. Given that we don't know whether this purely a theoretical situation, I'd prefer to err on the side of caution and include this disclaimer in case it becomes practical. It can be useful if a developer is debugging this situation and fails to recognize that adjusting the size of their submissions is an easy solution.

That said, I'm comfortable with it not being included if there is specific reason not to. The gas limit constraint is pertinent to almost every situation where an array is passed in as an argument, so this is no unique scenario. Moreover, the worst case scenario is lost gas / a wasted tx, not a harmful state change, so not including this comment doesn't put developers at extreme risk.

Either way, this is a very small change, and impact of either decision isn't that big -- I recommend including the comment, but am ok with not including.

0xrajath (approved):

src/ServiceManagerBase.sol
0xrajath: This would most likely be a large theoretical limit. But it's good to be explicit.

stevennevins: sgtm i had meant explicitly documenting the array size if users were likely to experience it. i'm fine with the comment as is since it doesn't seem like users will face it in practical scenarios so we're aligned there

PR #350, opened 2025-01-07
registry coordinator refactor

ypatil12:
- Use a single `_registerOperator` function, mimicking how there is a single `_deregisterOperator` internal function
- Remove M2Quorum storage, check if the operatorSet is M2 or Slashing by querying the `allocationManager`
- Allow parallel M2 & slashing registrations. Only cutover when `becomeOperatorSetAVS` is called

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -155,6 +155,9 @@ contract RegistryCoordinator is
          */
         bytes32 operatorId = _getOrCreateOperatorId(msg.sender, params);
 
+        // Validate that the quorums are M2 quorums and not operatorSets in the AllocationManager
+        _validateM2Quorums(quorumNumbers);
+
         // Register the operator in each of the registry contracts and update the operator's
         // quorum bitmap and registration status
         uint32[] memory numOperatorsPerQuorum = _registerOperator({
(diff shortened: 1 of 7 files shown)


src/RegistryCoordinator.sol
```diff
@@ -455,6 +466,15 @@ contract RegistryCoordinator is
         emit OperatorSocketUpdate(_operatorInfo[msg.sender].operatorId, socket);
     }
 
+    /**
+     * @notice Become an operator set AVS, preventing future M2, AVSDirectory-based, registrations
+     * @dev This function should only be called once sufficient stake operatorSet registrations in core has been acquired
+     */
+    function becomeOperatorSetAVS() external onlyOwner {
+        // Enable operator sets mode
+        isOperatorSetAVS = true;
```
8sunyuan, line 475: Mentioned this in the original PR but imo this should be the default setting and have this set to true in the `initialize` function

src/RegistryCoordinator.sol
```diff
@@ -255,58 +276,36 @@ contract RegistryCoordinator is
...
+    /**
+     * @notice Registers the `operator` for one ore more operatorSets. If any operatorSet exceeds its maximum
+     * operator capacity after the operator is registered, this method will fail.
+     * @param operator the operator to register
+     * @param operatorSetIds is the list of operatorSetIds to be registered for
+     * @param data Custom data containing The G1 & G2 public keys of the operator, and a signature proving their ownership
+     * @dev This function is an AllocationManager-based registration where the msg.sender is the AllocationManager
+     * @dev `data` is  ignored if the caller has previously registered a public key
+     */
     function registerOperator(
```
8sunyuan, line 303: Note: for this function, since its being called by the ALM `!_validateM2Quorums(quorumNumbers);` is not needed as its already validated by the ALM

8sunyuan: So for registration, M2 quorums can only be registered for through the legacy interface and OperatorSet quorums can only be registered through the new ALM callback. Similarly for deregistering, there are 2 separate interfaces for the respective quorum types.

PR #353, opened 2025-01-08
feat: operator set rewards

0xClandestine:


test/unit/ServiceManagerBase.t.sol
```diff
@@ -217,12 +224,29 @@ contract ServiceManagerBase_UnitTests is MockAVSDeployer, IServiceManagerBaseEve
         return timestamp1 > timestamp2 ? timestamp1 : timestamp2;
     }
 
+    // function test_setRewardsInitiator() public {
```
0xClandestine, line 227: TODO: uncomment

src/ServiceManagerBase.sol
0xrajath, line 1: there's an ECDSAServiceManager too that you'll need to update.

PR #354, opened 2025-01-09
feat: avs registrar

ypatil12:
This PR creates a new contract, the `Registrar`, which handles all interactions with the AllocationManager. Summary of changes:

## Registar
- Handles registration calls from the `AllocationManager`. These calls can include churn, which is handled by adding a `RegistrationType`
- All other functionality/interfaces match with the `RegistryCoordinator`
- _Note_: Due to deregistrations from the ALM not failing, if the `operatorSetIds` are not ordered then state will not update in middleware but will in core. The AVS must eject the operator manually if this is the case

## ServiceManager
- Remove all pass throughs into ALM for operatorSet updates. Relies on UAM for communication with all registries. 
- `PermissionController` integration via adding a pending admin

### `RegistryCoordinator`
- Only handles `AVSDirectory` based registration
- Can no longer create quorums once upgraded
- Is unaware of operatorSets or M2 state
### `StakeRegistry`/`IndexRegistry`
- Upgraded to handle registration calls from `RegistryCoordinator` and `Registrar`
- `StakeRegistry` makes calls to the ALM through UAM for adding/removing strats from an opSet

## Upgrade Process
The upgrade to middleware proceeds in 3 phases:
### Phase 1 - Begin Receiving ALM registrations, Operations continue as normal on M2
1. Upgrade all contracts & initiate UAM transactions
2. `RegistryCoordinator` & `Registrar` processes registrations/deregistrations in separate states. Quorums continue to be non-overlapping since the `RegistryCoordinator` can no longer create quorums and the `Registrar` initiates its quorumCount from the `RegistryCoordinator's`. 
### Phase 2 - Deprecate `RegistryCoordinator`. Promote `Registrar` to primary in all off-chain software. BLSVerification occurs against the `Registrar`
1. Upgrade all contracts to deprecate `RegistryCoordinator`. BLSSigChecker is updated to read from the `Registrar`
### Phase 3 - MMGA
1. Use `Registrar` and UAM as a starting point


8sunyuan:
This looks really nice and I think I am leaning more towards this solution. I notice some of the key differences between the original PR are the following:
1. Removing `isOperatorSetAVS` and `isM2Quorum`. There are 2 separate bitmaps in the respective contracts for M2 quorums and OperatorSet quorums. Allowing for full deregistration in the AVSD if all M2 quorums are deregistered for an operator. This was unaccounted for if they were registered for some operatorSets.
2. Allows for both legacy and operatorSets registration, allowing for a gradual migration of operators into the new quorums and registration flow. Old legacy interfaces on RegistryCoordinator are not blocked by setting `enableOperatorSets` and this contract is simply deprecated by the AVS at some later date. Mirroring how the AVSD is to be deprecated in the core protocol as well.
3. Logic is much cleaner and surely saves a lot of bytecode too

PR #357, opened 2025-01-15
refactor: slashing UAM

8sunyuan:
### Description

Adding UAM support to middleware. The `PermissionController` contract in EigenLayer allows for AVSs to configure specific (target address, function selector) tuples to be called by a designated address. This implies that calls to the EigenLayer core contracts are no longer required to be proxied through the ServiceManager. Example configuration 
The ServiceManager contract is both the address identifier for the AVS and by default the 'Admin' who can configure such targets and selectors. 

The changes to the core contracts implies the following configurations on the `PermissionController` contract:
1. `AllocationManager.createOperatorSets` - set to the RegistryCoordinator as creating a quorum involves calling the AllocationManager to create an operatorSet.
2. `AllocationManager.addStrategiesToOperatorSet` - set to the StakeRegistry, adding Strategies to a quorum will also forward a call to the AllocationManager
3. `AllocationManager.removeStrategiesFromOperatorSet` - set to the StakeRegistry, similar to adding Strategies.
4. `AllocationManager.slashOperator` - set to the Slasher contract, see changes to `SlasherBase.sol`

#### Additional Changes
- fixed `StakeRegistry._getSlashableStakePerStrategy` to use blocks instead of timestamps to align with the AllocationManager which uses blocks
- removed Slasher proposal delay originally implemented in the ServiceManagerBase contract. However the ServiceManager could instantly change the target selector for the `slashOperator` and effectively the Slasher contract. For better safety measures around upgrades, this proposal delay could be emulated by setting an [AccessManager](https://docs.openzeppelin.com/contracts/5.x/api/access#AccessManager) contract as the new admin for the AVS.


src/ServiceManagerBase.sol
```diff
@@ -40,45 +39,64 @@ abstract contract ServiceManagerBase is ServiceManagerBaseStorage {
...
         )
     {
         _disableInitializers();
     }
 
     function __ServiceManagerBase_init(
         address initialOwner,
-        address _rewardsInitiator,
-        address _slasher
+        address _rewardsInitiator
```
stevennevins, line 63: Are we leaving this as is because of rewards v2 audit?

8sunyuan: This is actually for RewardsV1 which we need backwards compatability for. `getRestakeableStrategies` and `getOperatorRestakedStrategies` are also needed for rewards V1

stevennevins: makes sense

ypatil12 (commented):
This looks good, do we also want to inherit `AccessManaged` and show examples in tests, or do in a separate PR?

8sunyuan:
> This looks good, do we also want to inherit `AccessManaged` and show examples in tests, or do in a separate PR?

Thinking of a followup PR so we can merge changes in right now

stevennevins (approved):

PR #360, opened 2025-01-21
feat: add register with churn type

stevennevins:


ypatil12 (approved):

ypatil12 (commented):
Realized `_deregisterOperator` doesn't deregister from ALM

src/RegistryCoordinator.sol
```diff
@@ -293,38 +293,114 @@ contract RegistryCoordinator is
...
+                    totalQuorumStake: results.totalStakes[i],
+                    newOperator: operator,
+                    newOperatorStake: results.operatorStakes[i],
+                    kickParams: operatorKickParams[i],
+                    setParams: operatorSetParams
+                });
+
+                bytes memory singleQuorumNumber = new bytes(1);
+                singleQuorumNumber[0] = quorumNumbers[i];
+                _deregisterOperator(operatorKickParams[i].operator, singleQuorumNumber);
```
ypatil12, line 406: This function doesn't deregister from the ALM

ypatil12 (changes requested):

src/RegistryCoordinator.sol
stevennevins: ahh good call, will add that

should resolve the ejction via ALM as well

src/RegistryCoordinator.sol
```diff
@@ -477,10 +556,42 @@ contract RegistryCoordinator is
...
         }
     }
 
+    /**
+     * @notice Helper function to handle operator set deregistration for non-M2 quorums
+     * @param operator The operator to deregister
+     * @param quorumNumbers The quorum numbers to check
+     */
+    function _handleOperatorSetDeregistration(address operator, bytes memory quorumNumbers) internal {
+        uint32[] memory nonM2OperatorSetIds = new uint32[](quorumNumbers.length);
```
ypatil12, line 570: I feel like we can change this with the hooks we have in Michael's PR, but for now this works

ypatil12 (approved):

src/RegistryCoordinator.sol
stevennevins: yeah def

PR #361, opened 2025-01-21
refactor: slashing registry coordinator

8sunyuan:
Changes:
- removed `isUsingOperatorSets()` and just read storage bool directly
- setting isOperatorSetAVS to default to true in `initialize`
- Moved all logic to `SlashingRegistryCoordinator`, `RegistryCoordinator` now only contains the legacy M2 interfaces `registerOperator`, `registerOperatorWithChurn`, `deregisterOperator` as well as `enableOperatorSets` for migration.
- `registerOperator` hook call from the ALM was missing the maxOperatorCount check in the NORMAL RegistrationType
- Cleaned up some internal interfaces and moved the `_getOrCreateOperatorId` call into `registerOperator`.
- renamed deregistration function calling back to ALM as `_forceDeregisterOperator` for clarity, also added this to `ejectOperator` as it was missing there.

Register/Deregister Hook changes:
- `_beforeRegisterOperator` Hook to allow for any pre-register logic in `_registerOperator`
- `_afterRegisterOperator` Hook to allow for any post-register logic in `_registerOperator`
- `_beforeDeregisterOperator` Hook to allow for any pre-deregister logic in `_deregisterOperator`
- `_afterDeregisterOperator` Hook to allow for any post-deregister logic in `_deregisterOperator`
    - `RegistryCoordinator overrides this to call the ServiceManager to deregister if the operator is no quorums anymore.
- Because the call to the ServiceManager for registering operators requires a signature, this was not used in a hook and was added to the external functions themselves in `RegistryCoordinator.registerOperator` and `RegistryCoordinator.registerOperatorWithChurn`

Note: will run a `forge fmt src` after finished review


src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1054 @@
...
+    }
+
+    function registerOperator(
+        address operator,
+        uint32[] memory operatorSetIds,
+        bytes calldata data
+    ) external override onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
+        require(isOperatorSetAVS, OperatorSetsNotEnabled());
+        for (uint256 i = 0; i < operatorSetIds.length; i++) {
+            require(!isM2Quorum[uint8(operatorSetIds[i])], OperatorSetsNotSupported());
```
ypatil12: Aren't we not an M2Quorum by default?

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1054 @@
...
+        }
+        _checkAllocationManager();
```
ypatil12: Saw this earlier, wonder if we can make this a modifier

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1054 @@
...
+         * Update operator's bitmap, socket, and status. Only update operatorInfo if needed:
+         * if we're `REGISTERED`, the operatorId and status are already correct.
+         */
+        _updateOperatorBitmap({operatorId: operatorId, newBitmap: newBitmap});
+
+        emit OperatorSocketUpdate(operatorId, socket);
+
+        // If the operator wasn't registered for any quorums, update their status
+        // and register them with this AVS in EigenLayer core (DelegationManager)
+        if (_operatorInfo[operator].status != OperatorStatus.REGISTERED) {
```
ypatil12: Is it possible to bring this logic up to the RegistryCoordinator since its dealing with the ServiceManager & has M2 logic?

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1054 @@
...
+
+        // Update operator's bitmap and status
+        _updateOperatorBitmap({operatorId: operatorId, newBitmap: newBitmap});
+
+        // If the operator is no longer registered for any quorums, update their status and deregister
+        // them from the AVS via the EigenLayer core contracts
+        if (newBitmap.isEmpty()) {
+            operatorInfo.status = OperatorStatus.DEREGISTERED;
+            serviceManager.deregisterOperatorFromAVS(operator);
+            emit OperatorDeregistered(operator, operatorId);
```
ypatil12: Wonder if we can leave the serviceManager logic to the normal `RegistryCoordinator`. Thinking that a new AVS wouldn't even have a serviceManager (set it to the 0 address)

src/SlashingRegistryCoordinator.sol
8sunyuan: We are, only the inheriting `RegistryCoordinator` makes use of this check. Perhaps we can remove this check here and only have it in `RegistryCoordinator` and calling the super implementation

src/SlashingRegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/b5ed8be4f2d9153d7ee750af887b46ad32c095d3
Removed all references of ServiceManager here

src/SlashingRegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/b5ed8be4f2d9153d7ee750af887b46ad32c095d3 Done, used some hooks to override

src/SlashingRegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/b5ed8be4f2d9153d7ee750af887b46ad32c095d3

src/RegistryCoordinator.sol
```diff
@@ -132,23 +51,14 @@ contract RegistryCoordinator is
...
-     */
+    /// @inheritdoc IRegistryCoordinator
     function registerOperator(
         bytes memory quorumNumbers,
         string memory socket,
         IBLSApkRegistry.PubkeyRegistrationParams memory params,
         SignatureWithSaltAndExpiry memory operatorSignature
     ) external onlyWhenNotPaused(PAUSED_REGISTER_OPERATOR) {
-        require(!isUsingOperatorSets(), OperatorSetsEnabled());
+        require(!isOperatorSetAVS, OperatorSetsEnabled());
```
ypatil12: Need to handle parallel registrations, assume we're doing so in a later commit?

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1012 @@
...
+    /**
+     * @notice Register the operator for one or more quorums. This method updates the
+     * operator's quorum bitmap, socket, and status, then registers them with each registry.
+     */
+    function _registerOperatorToOperatorSet(
+        address operator,
+        bytes32 operatorId,
+        bytes memory quorumNumbers,
+        string memory socket
+    ) internal virtual returns (RegisterResults memory results) {
```
ypatil12, line 455: All of the logic here is basically the same as `_registerOperator` in the `RegistryCoordinator`. I think we can combine these. Th only thing we'd need to add is in the `afterRegister` hook on the `RegistryCoordinator` is to check the status and if registered then we call the SM.

src/SlashingRegistryCoordinator.sol
ypatil12: I think it's fine to assume all calls from the ALM are for "new" quorums and then remove this check

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1012 @@
...
+            stakeRegistry.registerOperator(operator, operatorId, quorumNumbers);
+        results.numOperatorsPerQuorum = indexRegistry.registerOperator(operatorId, quorumNumbers);
+
+        // call hook to allow for any post-register logic
+        _afterRegisterOperator(operator, operatorId, quorumNumbers, newBitmap);
+
+        return results;
+    }
+
+    function _registerOperatorWithChurn(
```
ypatil12, line 500: I think we can also consolidate logic with the `RegistryCoordinator` for this function. 

`RegistryCoordinator` just has to properly handle registration back to AVSD

ypatil12 (commented):
Organizational comments

src/SlashingRegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/c462c88593e6f0ce33c74e82e51ae8925e9a83fa
reusing the churn logic here

src/SlashingRegistryCoordinator.sol
8sunyuan: Removed the `_registerOperator` function in RegistryCoordinator and reusing the logic. Rather than using the hook, I just added the SM call to the `RegistryCoordinator` itself since it requires an operatorSignature to be passed into the SM

src/SlashingRegistryCoordinator.sol
```diff
@@ -0,0 +1,1097 @@
...
+            });
+        } else {
+            revert InvalidRegistrationType();
+        }
+
+        // If the operator wasn't registered for any quorums, update their status
+        // and register them with this AVS in EigenLayer core (DelegationManager)
+        if (_operatorInfo[operator].status != OperatorStatus.REGISTERED) {
+            _operatorInfo[operator] = OperatorInfo(operatorId, OperatorStatus.REGISTERED);
+            emit OperatorRegistered(msg.sender, operatorId);
```
stevennevins: msg.sender -> operator

msg.sender is the ALM

stevennevins (approved):

src/RegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/57876ae5ff47d89a1a756147047a6cc38dd27be8

src/SlashingRegistryCoordinator.sol
8sunyuan: nice catch, fixed here https://github.com/Layr-Labs/eigenlayer-middleware/pull/361/commits/66fe71a645d4548ce7e5e6707aecdd6cb8928a1d

PR #379, opened 2025-02-01
refactor: slashing

gpsanant:
main changes

- slashing registry coordinator doesn't know about m2
- slashing registry coordinator added before/after quorum hooks
- added a `checkMaxOperatorCount` to registration so less duplicated logic in registry coordinator
- removed owner from slashing registry coordinator interface, now just cast and use
- stake registry cleanup
- refactor AVS sync to make a single call to the core contracts
- lazy migration to operator sets upon the next created quorum, removal of `enableOperatorSets` interface

### DIFF
--- a/src/BLSSignatureChecker.sol
+++ b/src/BLSSignatureChecker.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+
 import {BitmapUtils} from "./libraries/BitmapUtils.sol";
 import {BN254} from "./libraries/BN254.sol";
 
@@ -18,7 +20,7 @@ contract BLSSignatureChecker is BLSSignatureCheckerStorage {
     /// MODIFIERS
 
     modifier onlyCoordinatorOwner() {
-        require(msg.sender == registryCoordinator.owner(), OnlyRegistryCoordinatorOwner());
+        require(msg.sender == Ownable(address(registryCoordinator)).owner(), OnlyRegistryCoordinatorOwner());
         _;
     }
 
--- a/src/BLSSignatureChecker.sol
+++ b/src/BLSSignatureChecker.sol
@@ -20,7 +20,10 @@ contract BLSSignatureChecker is BLSSignatureCheckerStorage {
     /// MODIFIERS
 
     modifier onlyCoordinatorOwner() {
-        require(msg.sender == Ownable(address(registryCoordinator)).owner(), OnlyRegistryCoordinatorOwner());
+        require(
+            msg.sender == Ownable(address(registryCoordinator)).owner(),
+            OnlyRegistryCoordinatorOwner()
+        );
         _;
     }
 
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -15,6 +15,7 @@ import {BitmapUtils} from "./libraries/BitmapUtils.sol";
 import {SlashingRegistryCoordinator} from "./SlashingRegistryCoordinator.sol";
 import {ISlashingRegistryCoordinator} from "./interfaces/ISlashingRegistryCoordinator.sol";
 import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import {RegistryCoordinatorStorage} from "./RegistryCoordinatorStorage.sol";
 
 /**
  * @title A `RegistryCoordinator` that has three registries:
@@ -24,11 +25,8 @@ import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/Ownabl
  *
  * @author Layr Labs, Inc.
  */
-contract RegistryCoordinator is IRegistryCoordinator, SlashingRegistryCoordinator {
+contract RegistryCoordinator is RegistryCoordinatorStorage {
     using BitmapUtils for *;
-
-    /// @notice the ServiceManager for this AVS, which forwards calls onto EigenLayer's core contracts
-    IServiceManager public immutable serviceManager;
 
     constructor(
         IServiceManager _serviceManager,
@@ -39,17 +37,16 @@ contract RegistryCoordinator is IRegistryCoordinator, SlashingRegistryCoordinato
         IAllocationManager _allocationManager,
         IPauserRegistry _pauserRegistry
     )
-        SlashingRegistryCoordinator(
+        RegistryCoordinatorStorage(
+            _serviceManager,
             _stakeRegistry,
             _blsApkRegistry,
             _indexRegistry,
             _socketRegistry,
             _allocationManager,
             _pauserRegistry
-        )
-    {
-        serviceManager = _serviceManager;
-    }
+        ) 
+    {}
 
     /**
      *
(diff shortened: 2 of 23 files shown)


test/unit/RegistryCoordinatorUnit.t.sol
```diff
@@ -1947,91 +1947,91 @@ contract RegistryCoordinatorUnitTests_RegisterOperatorWithChurn is RegistryCoord
...
-        address[] memory operatorsToUpdate = new address[](1);
-        operatorsToUpdate[0] = defaultOperator;
-
-        // force a staticcall to the `updateOperators` function -- this should *pass* because the call should be a strict no-op!
-        (bool success,) = address(registryCoordinator).staticcall(
-            abi.encodeWithSignature("updateOperators(address[])", operatorsToUpdate)
-        );
-        require(success, "staticcall failed!");
-    }
+    // function test_updateOperators_revert_paused() public {
```
gpsanant: delete?

src/SlashingRegistryCoordinator.sol
```diff
@@ -229,45 +222,16 @@ contract SlashingRegistryCoordinator is
...
         _deregisterOperator(operator, quorumNumbers);
     }
 
-    /**
-     * @notice Updates the StakeRegistry's view of one or more operators' stakes. If any operator
-     * is found to be below the minimum stake for the quorum, they are deregistered.
-     * @dev stakes are queried from the Eigenlayer core DelegationManager contract
-     * @param operators a list of operator addresses to update
-     */
-    function updateOperators(
```
gpsanant, line 244: we don't need to support this

stevennevins: it should stay for backward compatibility imho

test/unit/RegistryCoordinatorUnit.t.sol
stevennevins: lets add back if we're on board with keeping the function

src/SlashingRegistryCoordinatorStorage.sol
```diff
@@ -118,5 +106,5 @@ abstract contract SlashingRegistryCoordinatorStorage is ISlashingRegistryCoordin
 
     // storage gap for upgradeability
     // slither-disable-next-line shadowing-state
-    uint256[37] private __GAP;
+    uint256[40] private __GAP;
```
8sunyuan: Isn't this 2 slots in gap? Since the bools in storage are packed in 1 slot

src/RegistryCoordinatorStorage.sol
```diff
@@ -0,0 +1,69 @@
...
+            _indexRegistry,
+            _socketRegistry,
+            _allocationManager,
+            _pauserRegistry
+        )
+    {
+        serviceManager = _serviceManager;
+    }
+
+    uint256[47] private __GAP;
```
8sunyuan: should be 48

src/RegistryCoordinator.sol
```diff
@@ -64,44 +61,28 @@ contract RegistryCoordinator is IRegistryCoordinator, SlashingRegistryCoordinato
...
-         * (operatorId) is fetched instead
-         */
-        bytes32 operatorId = _getOrCreateOperatorId(msg.sender, params);
-
-        // Register the operator in each of the registry contracts and update the operator's
-        // quorum bitmap and registration status
-        uint32[] memory numOperatorsPerQuorum = _registerOperator({
+        require(!isM2QuorumRegistrationDisabled, M2QuorumRegistrationIsDisabled());
+        require(
+            quorumNumbers.orderedBytesArrayToBitmap().isSubsetOf(m2QuorumBitmap),
```
8sunyuan: Good catch here with this check, this was missed once we added parallel registration for Opsets and legacy quorums.
This require needs to be wrapped with `if (operatorSetsEnabled) {` since for existing AVSs who upgrade, this bitmap in storage will not be initialized if remain on legacy mode.

8sunyuan: Same for the other register interface

src/SlashingRegistryCoordinatorStorage.sol
gpsanant: bool
bool
address
uint256

=>

address

that's just 1 slot saved, net? so 38?

src/RegistryCoordinator.sol
gpsanant: instead, is it ok to just check that operatorSets are enabled when disabling m2 quorum registration?

gpsanant: https://github.com/Layr-Labs/eigenlayer-middleware/pull/379/commits/8ae606e7b876e4501a66cd5d326880405761c3bf

src/RegistryCoordinatorStorage.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-middleware/pull/379/commits/8ae606e7b876e4501a66cd5d326880405761c3bf

src/SlashingRegistryCoordinatorStorage.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-middleware/pull/379/commits/8ae606e7b876e4501a66cd5d326880405761c3bf

src/RegistryCoordinator.sol
8sunyuan: I think something like 
```
require(
    !operatorSetsEnabled || quorumNumbers.orderedBytesArrayToBitmap().isSubsetOf(m2QuorumBitmap),
    OnlyM2QuorumsAllowed()
);
```

So if operatorSets are not enabled, then we don't need to check the quorumNumbers because no opSet quorums exist. Otherwise we have to do the explicit check

src/SlashingRegistryCoordinatorStorage.sol
8sunyuan: 1 slot from the 2 bools. `M2quorumBitmap` removed which is another so from 37 -> 39

src/RegistryCoordinator.sol
```diff
@@ -174,20 +150,47 @@ contract RegistryCoordinator is IRegistryCoordinator, SlashingRegistryCoordinato
...
+            // call the parent _forceDeregisterOperator function for operator sets quorums
+            super._forceDeregisterOperator(operator, operatorSetBitmap.bitmapToBytesArray());
+        }
+    }
+
+    /// @dev Hook to prevent any new quorums from being created if operator sets are not enabled
+    function _beforeCreateQuorum(
+        uint8
+    ) internal virtual override {
+        require(operatorSetsEnabled, OperatorSetsNotEnabled());
```
8sunyuan: This is blocking creation of new quorums if you are a legacy AVS who has upgraded but haven't enabledOperatorSets.

src/SlashingRegistryCoordinator.sol
```diff
@@ -441,12 +413,29 @@ contract SlashingRegistryCoordinator is
 
         emit OperatorSocketUpdate(operatorId, socket);
```
8sunyuan:
```suggestion
        _setOperatorSocket(operatorId, socket);
```
Needs to be set in SocketRegistry, can you include this fix in this PR

src/RegistryCoordinator.sol
8sunyuan: > should we get rid of enableOperatorSets() and just run that logic lazily upon the first operatorSet quorum created?

I think this is actually the better way to go. As an existing AVS who upgrades and decides to create a new quorum, it should be expected that an OperatorSet quorum should be created.

stevennevins: Yeah i think that makes logical sense

src/SlashingRegistryCoordinatorStorage.sol
gpsanant: slot1
- bool
- bool
- address

slot2
- uint256


=>

slot1
- address


or am i being stupid

8sunyuan: Ah I'm being stupid. Reread it again and you are right

src/RegistryCoordinator.sol
gpsanant: is it necessary or gas optimization?

src/RegistryCoordinator.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-middleware/pull/379/commits/333cdae5c98324b88a10dde61c2cf3955d782c3e

src/RegistryCoordinator.sol
gpsanant: https://github.com/Layr-Labs/eigenlayer-middleware/pull/379/commits/333cdae5c98324b88a10dde61c2cf3955d782c3e ok?

8sunyuan: We have to do the explicit check otherwise the m2QuorumBitmap won't be initialized if you never emabledOperatorSets and you can't register any legacy quorums

src/RegistryCoordinator.sol
8sunyuan, line 1: As mentioned offline, think keeping `enableOperatorSets` interface is good to have so that M2 registration can be disabled without having to create a new quorum. Would allow an existing AVS to just deprecate an old RegistryCoordinator and deploy a new SlashingRegistryCoordinator if they wish as part of their onchain migration.

src/RegistryCoordinator.sol
```diff
@@ -218,6 +221,17 @@ contract RegistryCoordinator is RegistryCoordinatorStorage {
      *
      */
 
+    /// @dev Returns a bitmap with all bits set up to `quorumCount`. Used for bit-masking quorum numbers
+    /// and differentiating between operator sets and M2 quorums
+    function m2QuorumBitmap() public view returns (uint256) {
```
gpsanant, line 225: "avsDirectoryQuorum"?

src/RegistryCoordinator.sol
8sunyuan: Settled on removing the enableOperatorSets check on disableM2registrations call so above flow is possible

src/BLSSignatureChecker.sol
```diff
@@ -18,7 +20,10 @@ contract BLSSignatureChecker is BLSSignatureCheckerStorage {
     /// MODIFIERS
 
     modifier onlyCoordinatorOwner() {
-        require(msg.sender == registryCoordinator.owner(), OnlyRegistryCoordinatorOwner());
+        require(
+            msg.sender == Ownable(address(registryCoordinator)).owner(),
```
0xClandestine, line 24: Is RC interface inheriting IOwnable?

src/interfaces/IRegistryCoordinator.sol
```diff
@@ -107,11 +111,13 @@ interface IRegistryCoordinator is
     ) external;
 
     /**
-     * @notice Enables operator sets mode for the AVS. Once enabled, this cannot be disabled.
-     * @dev When enabled, all existing quorums are marked as M2 quorums and future quorums must be explicitly
-     * created as either M2 or operator set quorums.
```
0xClandestine, line 112: inherit doc, i think?

0xClandestine (approved):

src/BLSSignatureChecker.sol
8sunyuan: just the `owner()` interface is needed here but the `SlashingRegistryCoordinator` does inherit `OwnableUpgradeable`

stevennevins (approved):

PR #382, opened 2025-02-04
refactor: ci

0xClandestine:


stevennevins (approved):

test/unit/BitmapUtils.t.sol
```diff
@@ -216,10 +237,22 @@ contract BitmapUtilsUnitTests_bytesArrayToBitmap is BitmapUtilsUnitTests {
...
     function testFuzz_BytesArrayToBitmapToBytesArray_OrderedVersion_Yul(
-        bytes memory originalBytesArray
-    ) public {
-        // filter down to only ordered inputs
-        cheats.assume(bitmapUtilsWrapper.isArrayStrictlyAscendingOrdered(originalBytesArray));
+        uint8 a,
+        uint8 b,
+        uint8 c
+    ) public view {
+        a = uint8(bound(a, 0, 31));
```
stevennevins, line 244: are you sure this was capped at 31 bytes before?

0xClandestine: it's capping each byte at 31, just an arbitrary value, could be passed in as a fuzz param.

PR #386, opened 2025-02-06
Update submodule lib/eigenlayer-contracts to rewardsv2.1

Sidu28:


Sidu28:
@stevennevins looks like the coverage metric is not meeting a threshold - how important is that

stevennevins (approved):

PR #388, opened 2025-02-07
test: integration with slashingRegistryCoordinator

8sunyuan:
Edge case fixed:
The AllocationManager has a try/catch in the `deregisterFromOperatorSets` implementation. This was put in place to allow operators to force deregister from an AVS. However this can lead to a weird state where an operator is registered in AVS contracts but deregistered from EigenLayer. Checks for this are added for forceDeregistration by the AVS to avoid reverts. In addition, AVSSync (`updateOperatorsForQuorum`, `updateOperator`) will now also include these operators to be force deregistered.

Refactored integration tests to only use the `SlashingRegistryCoordinator` and defines a `OperatorSetUser` contract.
The `SlashingRegistryCoordinator` and the Registry contracts do not depend on the ServiceManager base contract.
Interactions with the core protocol are integrated with UAM and setting the `avsAccountIdentifier` as the unique account identifier address for the AVS. For existing AVSs, this will be the address of your ServiceManager contract.

TODOs, could be implemented in followup PR
1. integration tests with the `RegistryCoordinator` and legacy quorums will require some fork integration testing refactor and testing the legacy interfaces. This is still a TODO that should be added.
2. Integration tests with slashing and allocations
    - Should test the following, allocate, register to AVS after meeting slashable stake threshold, slash operator, update stakes, expect operator to have been force deregistered
    - allocate, register to AVS meeting slashable stake threshold, create vetoeable slashing request, veto slash, expect operator magnitude and shares to be unchanged

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -2,8 +2,10 @@
 pragma solidity ^0.8.27;
 
 import {IPauserRegistry} from "eigenlayer-contracts/src/contracts/interfaces/IPauserRegistry.sol";
-import {IAllocationManager} from
-    "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
+import {
+    IAllocationManager,
+    OperatorSet
+} from "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
 import {IBLSApkRegistry, IBLSApkRegistryTypes} from "./interfaces/IBLSApkRegistry.sol";
 import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
 import {IIndexRegistry} from "./interfaces/IIndexRegistry.sol";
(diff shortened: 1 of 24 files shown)


src/SlashingRegistryCoordinator.sol
```diff
@@ -249,13 +249,24 @@ contract SlashingRegistryCoordinator is
                     singleOperator, singleOperatorId, quorumNumber
                 );
 
-                if (shouldBeDeregistered[0]) {
+                // whether the operator is registered in the core EigenLayer contract AllocationManager
+                bool registeredInCore = allocationManager.isMemberOfOperatorSet(
```
gpsanant: can we call this before to exit early?

gpsanant: need ths logic in avs sync too!

8sunyuan: What's the condition to exit early. If the bool is true or false we need to handle accordingly.

src/RegistryCoordinator.sol
```diff
@@ -200,6 +202,44 @@ contract RegistryCoordinator is RegistryCoordinatorStorage {
...
+        bytes32[] memory operatorIds,
+        uint8 quorumNumber
+    ) internal virtual override {
+        bytes memory singleQuorumNumber = new bytes(1);
+        singleQuorumNumber[0] = bytes1(quorumNumber);
+        bool[] memory doesNotMeetStakeThreshold =
+            stakeRegistry.updateOperatorsStake(operators, operatorIds, quorumNumber);
+        for (uint256 j = 0; j < operators.length; ++j) {
+            // whether the operator is registered in the core EigenLayer contract AllocationManager
+            bool registeredInCore = allocationManager.isMemberOfOperatorSet(
```
gpsanant: only do this call if it's an m2 quorum imo

src/RegistryCoordinator.sol
```diff
@@ -196,6 +202,54 @@ contract RegistryCoordinator is RegistryCoordinatorStorage {
...
+        address[] memory operators,
+        bytes32[] memory operatorIds,
+        uint8 quorumNumber
+    ) internal virtual override {
+        bytes memory singleQuorumNumber = new bytes(1);
+        singleQuorumNumber[0] = bytes1(quorumNumber);
+        bool[] memory doesNotMeetStakeThreshold =
+            stakeRegistry.updateOperatorsStake(operators, operatorIds, quorumNumber);
+
+        for (uint256 j = 0; j < operators.length; ++j) {
```
0xClandestine: j -> i

src/SlashingRegistryCoordinator.sol
```diff
@@ -596,6 +626,44 @@ contract SlashingRegistryCoordinator is
...
 
+    /**
+     * @dev Helper function to update operator stakes and deregister loiterers
+     * Loiterers are AVS registered operators who have force deregistered from the OperatorSet/quorum
+     * in the core EigenLayer contract AllocationManager but not deregistered from the OperatorSet/quorum
+     * in this contract. Potentially due to out of gas errors in the deregistration callback. This function
+     * will handle that edge case by deregistering the operator from the AVS if they are no longer registered
+     * in the AllocationManager.
+     */
+    function _updateStakesAndDeregisterLoiterers(
```
0xClandestine, line 640: interesting naming..

src/slashers/base/SlasherBase.sol
```diff
@@ -12,7 +11,7 @@ import {IStrategy} from "eigenlayer-contracts/src/contracts/interfaces/IStrategy
 /// @title SlasherBase
 /// @notice Base contract for implementing slashing functionality in EigenLayer middleware
 /// @dev Provides core slashing functionality and interfaces with EigenLayer's AllocationManager
-abstract contract SlasherBase is Initializable, SlasherStorage {
```
0xClandestine, line 15: Not sure if this contract is currently deployed to production environments. Just wanted to point out that if it is, these changes will alter its storage layout upon upgrade.

0xClandestine, line 15: Not sure if this contract is currently deployed to production environments. Just wanted to point out that if it is, these changes will alter its storage layout upon upgrade.

0xClandestine (changes requested):
Some nits, looks good otherwise.

src/slashers/base/SlasherBase.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/388/commits/fe5c9dc53a05ead8e20ea5c26495ffbd832087df
Actually looked like gap was off after changing to immutable so fixed that as well

src/RegistryCoordinator.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/commit/fe5c9dc53a05ead8e20ea5c26495ffbd832087df

0xClandestine (approved):

stevennevins (approved):

PR #394, opened 2025-02-11
feat: configurable veto delay

gpsanant:
makes the veto delay configurable and changes to blocks


stevennevins (approved):

stevennevins:
forge fmt before merge

PR #397, opened 2025-02-11
feat: move addresses to constructor

gpsanant:


stevennevins:
do we want to make these only changeable via upgrades?

afkbyte (approved):

gpsanant:
> do we want to make these only changeable via upgrades?

That's the way we do our other contracts, so I'd say yes.

PR #402, opened 2025-02-14
fix: handle eject operators correctly for post try-catch

stevennevins:
Changes:

- ejectOperator in SlashingRegistryCoordinator now only uses _forceDeregisterOperator, allowing AllocationManager to handle callbacks for deregistration

- RegistryCoordinator now handles M2 and non-M2 quorums separately:

- M2 quorums use _deregisterOperator

- Non-M2 quorums use _forceDeregisterOperator

- Added bitmap intersection support to filter quorum types properly

This change improves the ejection flow by ensuring proper deregistration handling through the AllocationManager while maintaining special handling for M2 quorums on ejection

### DIFF
--- a/src/SlashingRegistryCoordinator.sol
+++ b/src/SlashingRegistryCoordinator.sol
@@ -651,22 +651,9 @@ contract SlashingRegistryCoordinator is
         bool[] memory doesNotMeetStakeThreshold =
             stakeRegistry.updateOperatorsStake(operators, operatorIds, quorumNumber);
         for (uint256 j = 0; j < operators.length; ++j) {
-            // whether the operator is registered in the core EigenLayer contract AllocationManager
-            bool registeredInCore = allocationManager.isMemberOfOperatorSet(
-                operators[j], OperatorSet({avs: accountIdentifier, id: uint32(quorumNumber)})
-            );
-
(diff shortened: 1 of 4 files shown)


src/SlashingRegistryCoordinator.sol
8sunyuan, line 1: `_deregisterOperator` should have the `_forceDeregister` and `shouldForceDeregister` input removed and only be called in the callback function

I think the following functions have `_deregisterOperator` calls and need to be updated with `_forceDeregister`.
- `_registerOperatorWithChurn`
- `_updateStakesAndDeregisterLoiterers` (this can be cleaned up a bit and renamed)

stevennevins: Ahh good call. Will update for those as well

stevennevins: We can probably remove the loiterers concept now since there isn't a risk of the registration status being out sync?

And we can combine churn with ejection on the implementation side

8sunyuan: Yup no more loiterers :)

stevennevins: Updated @8sunyuan

PR #404, opened 2025-02-18
test: increase `ServiceManagerBase` coverage

tsnewnami:
This PR increases the test coverage of the `ServiceManagerBase` contract by adding two fuzzing tests for `setClaimerFor`

### DIFF
--- a/test/unit/ServiceManagerBase.t.sol
+++ b/test/unit/ServiceManagerBase.t.sol
@@ -928,4 +928,27 @@ contract ServiceManagerBase_UnitTests is MockAVSDeployer, IServiceManagerBaseEve
         cheats.expectRevert(IServiceManagerErrors.OnlyRegistryCoordinator.selector);
         serviceManager.deregisterOperatorFromOperatorSets(operator, operatorSetIds);
     }
+
+    function testFuzz_setClaimerFor_revert_notOwner(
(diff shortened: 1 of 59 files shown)


test/unit/ServiceManagerBase.t.sol
```diff
@@ -928,4 +928,27 @@ contract ServiceManagerBase_UnitTests is MockAVSDeployer, IServiceManagerBaseEve
...
+
+        cheats.expectRevert("Ownable: caller is not the owner");
+        cheats.prank(caller);
+        serviceManager.setClaimerFor(claimer);
+    }
+
+    function testFuzz_setClaimerFor(
+        address claimer
+    ) public filterFuzzedAddressInputs(claimer) {
+        cheats.expectCall(
```
stevennevins: We can leave the expectCall out to keep things concise

I would rather see an assertion on the state change after

and then another test where we test and assert the event emit was correct

tsnewnami: Good idea. Updated: https://github.com/Layr-Labs/eigenlayer-middleware/pull/404/commits/66409867e0581514f557089573d8253598d7af20

stevennevins: lgtm

PR #418, opened 2025-03-06
feat: operatorStateWithSocket

0x0aa0:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/OperatorStateRetriever.sol
+++ b/src/OperatorStateRetriever.sol
@@ -20,13 +20,6 @@ contract OperatorStateRetriever {
         uint96 stake;
     }
 
-    struct OperatorWithSocket {
-        address operator;
-        bytes32 operatorId;
-        uint96 stake;
-        string socket;
-    }
-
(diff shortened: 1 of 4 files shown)


pakim249CAL:
This runs into stack too deep errors because of the addition of a string in the OperatorWithSocket struct. Because strings are dynamically sized, a memory reference to this string takes up a stack slot. Maybe this can be fixed with --via-ir compilation, but I believe the better immediate solution is to simply not implement this and instead use what is already available. 

In particular, we can obtain the same results offchain by querying the already existing getOperatorState on the OperatorStateRetriever, and querying a getOperatorSocket function that calls the same function on the SocketRegistry.

So I propose that for this PR, we only just add a getOperatorSocket function to OperatorStateRetriever.sol

pakim249CAL:
https://github.com/Layr-Labs/eigenlayer-middleware/pull/428 Proposed fix for stack too deep here

hopeyen (approved):
lgtm! thanks

PR #421, opened 2025-03-07
docs: add `OperatorStateRetriever`

tsnewnami:
**Motivation:**

`OperatorStateRetriever` had no documentation

**Modifications:**

Added `OperatorStateRetriever.md`

**Result:**

Increase documentation coverage


stevennevins (approved):

stevennevins:
One issue that comes up with the contract is that people aren't clear how they're supposed to use it to get the inputs for the BLSSignatureChecker checkSignatures call.  We might want to include that as a specific item in the docs.  This can be a fast follow addition

PR #427, opened 2025-03-12
chore: bump openzeppelin 4.9.0

stevennevins:
Motivation:
The core dependency was updated to OZ 4.9.0 a while ago. Now, it uses a contract only in 4.9.0, making it incompatible with 4.7.1 in this repo. Updating to 4.9.0 results in a breaking change to the TransparentUpgradeableProxy interface. These changes need to be addressed to maintain compatibility with the core protocol.

Modifications:
Updated TransparentUpgradeableProxy usage:
Modified imports to include ITransparentUpgradeableProxy alongside TransparentUpgradeableProxy
Updated proxy upgrade calls to cast to ITransparentUpgradeableProxy instead of TransparentUpgradeableProxy
Fixed ERC20Mock constructor errors:
Updated ERC20Mock in test files to use the new constructor pattern
Removed parameters from ERC20Mock constructor calls as the new version no longer accepts name, symbol, initialAccount, and initialBalance parameters

Result:
After these changes, middleware is compatible with the future EigenLayer core v1.3.0

### DIFF
--- a/test/integration/IntegrationDeployer.t.sol
+++ b/test/integration/IntegrationDeployer.t.sol
@@ -7,7 +7,8 @@ import "forge-std/Test.sol";
 import "@openzeppelin/contracts/token/ERC20/presets/ERC20PresetFixedSupply.sol";
 import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
 import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
-import {ITransparentUpgradeableProxy} from "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+import {ITransparentUpgradeableProxy} from
+    "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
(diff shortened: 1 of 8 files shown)


8sunyuan:
Needs to update OZ upgradeable lib

test/integration/IntegrationDeployer.t.sol
```diff
@@ -423,17 +425,9 @@ abstract contract IntegrationDeployer is Test, IUserDeployer {
...
-            TransparentUpgradeableProxy(payable(address(registryCoordinator))),
-            address(registryCoordinatorImplementation),
-            abi.encodeWithSelector(
-                SlashingRegistryCoordinator.initialize.selector,
-                registryCoordinatorOwner,
-                churnApprover,
-                ejector,
-                0, /*initialPausedStatus*/
-                address(serviceManager) /* accountIdentifier */
-            )
```
8sunyuan: Not initializing the RegistryCoordinator?

8sunyuan (approved):

test/integration/IntegrationDeployer.t.sol
stevennevins: Fixed!

PR #428, opened 2025-03-13
refactor: return strings outside of struct to avoid stack too deep

pakim249CAL:
**Motivation:**

The existing PR has stack too deep in tests. Need to move strings outside of the nested struct. There are a few long term things that we can do that I will list here, but I don't think it is necessary immediately because this is just an off-chain view getter that can easily be changed out later so it's just for reference.

* For some reason, via-ir compilation is disabled. This can often help with stack too deep issues, but will require some changes in the code that would expand the scope of the PR too much.
* We can make a socket a fixed size to fix the issue. Right now, a socket is an arbitrary length string. To me right now, the way socket is set is totally opaque so I can't really suggest anything about HOW we can make it a fixed size yet.

**Modifications:**

Move strings outside the nested struct.

**Result:**

Tests will work!

### DIFF
--- a/src/OperatorStateRetriever.sol
+++ b/src/OperatorStateRetriever.sol
@@ -20,41 +20,36 @@ contract OperatorStateRetriever {
         uint96 stake;
     }
 
-    struct OperatorWithSocket {
-        address operator;
-        bytes32 operatorId;
-        uint96 stake;
-        string socket;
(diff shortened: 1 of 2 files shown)


0x0aa0 (approved):
lgtm!

PR #433, opened 2025-03-15
fix: resolve inheritance order and storage layout for upgraded rc

stevennevins:
**Motivation:**

EIP712 in 4.7.0 did not have any storage. In 4.9.0 it has 2 slots of storage which would have caused a storage layout shift based on it's location in the inheritance graph for RegistryCoordinators going through an upgrade

**Modifications:**

Reorder inheritance, switch to the upgradeable version, and fix build issues that arose from the updates by adding the constructor params to a struct

**Result:**

Fixed the storage layout issue that would have arisen


tsnewnami (approved):

src/RegistryCoordinatorStorage.sol
```diff
@@ -11,12 +11,35 @@ import {IServiceManager} from "./interfaces/IServiceManager.sol";
...
+    /**
+     * @notice Parameters for initializing SlashingRegistryCoordinator
+     * @param stakeRegistry The StakeRegistry contract that keeps track of operators' stakes
+     * @param blsApkRegistry The BLSApkRegistry contract that keeps track of operators' BLS public keys
+     * @param indexRegistry The IndexRegistry contract that keeps track of ordered operator lists
+     * @param socketRegistry The SocketRegistry contract that keeps track of operators' sockets
+     * @param allocationManager The AllocationManager contract for operator set management
+     * @param pauserRegistry The PauserRegistry contract for pausing functionality
+     */
+    struct SlashingRegistryParams {
```
8sunyuan: Where are these params structs being used and why are they defined here?

src/RegistryCoordinator.sol
```diff
@@ -271,9 +265,9 @@ contract RegistryCoordinator is RegistryCoordinatorStorage {
 
     /// @notice Returns true if the quorum number is an M2 quorum
     /// @dev We use bitwise and to check if the quorum number is an M2 quorum
-    function _isM2Quorum(
+    function isM2Quorum(
         uint8 quorumNumber
-    ) internal view returns (bool) {
+    ) public view returns (bool) {
```
8sunyuan: Lets add to interface if this is public

foundry.toml
```diff
@@ -26,15 +26,15 @@
...
     via_ir = false
     # Whether or not to enable the Solidity optimizer.
     optimizer = true
-    # The number of runs specifies roughly how often each opcode of the deployed code will be executed 
-    # across the life-time of the contract. This means it is a trade-off parameter between code size (deploy cost) 
+    # The number of runs specifies roughly how often each opcode of the deployed code will be executed
+    # across the life-time of the contract. This means it is a trade-off parameter between code size (deploy cost)
     # and code execution cost (cost after deployment).
-    optimizer_runs = 200
+    optimizer_runs = 100
```
8sunyuan, line 37: I believe core repo is on 200 runs? perhaps keep as consistent so that contract sizes are as expected

src/SlashingRegistryCoordinator.sol
```diff
@@ -40,12 +41,12 @@ import {SlashingRegistryCoordinatorStorage} from "./SlashingRegistryCoordinatorS
...
  */
 contract SlashingRegistryCoordinator is
-    EIP712,
     Initializable,
     Pausable,
     OwnableUpgradeable,
     SlashingRegistryCoordinatorStorage,
-    ISignatureUtils
+    ISignatureUtils,
+    EIP712Upgradeable
```
8sunyuan: Would this cause issues for any AVSs using the RegistryCoordinator contract and inheriting from it with additional storage?
Needs to be a very clear breaking change

foundry.toml
stevennevins: it's a hack to get below the code size limit unfortunately

i'm not able to find a great longer team solution unfortunatley without significant rework

src/SlashingRegistryCoordinator.sol
stevennevins: yeah this is what i mentioned on standup

src/RegistryCoordinatorStorage.sol
stevennevins: they're being used to resolve a stack too deep issue in the constructor

src/RegistryCoordinator.sol
stevennevins: it's already in the interface.I just combined the two functions that exists to only be a public function instead of an internal and an external

src/RegistryCoordinatorStorage.sol
8sunyuan: Gotcha, nit is just to have it in the interface like the other structs but I get that its more viewable here in the storage file

stevennevins: Screenshot 2025-03-14 at 5 15 36 PM

stevennevins: sg

8sunyuan (approved):

PR #411, opened 2025-03-01
feat: Add operator set support for ECDSA stake registry

jaxxjj:
## Motivation:

With the slashing release upcoming, the current ECDSAStakeRegistry has not yet been updated to support operator sets. The system currently only supports M2 quorum registration, creating a need for a more flexible and comprehensive registration mechanism while maintaining backward compatibility.

### ECDSAStakeRegistry
- Renamed registration functions:
  - `registerOperatorWithSignature` → `registerOperatorM2Quorum`
  - `deregisterOperator` → `deregisterOperatorM2Quorum`
- Added operator set callbacks:
  - `onOperatorSetRegistered` - handles AllocationManager registration
  - `onOperatorSetDeregistered` - handles AllocationManager deregistration
- Added operator set management:
  - `getCurrentOperatorSetIds` - retrieves current operator set IDs
  - `getQuorumWeight` - calculates quorum weight
  - `getOperatorSetWeight` - calculates operator set weight
  - `operatorRegisteredOnAVSDirectory` - checks M2 quorum registration
  - `operatorRegisteredOnCurrentOperatorSets` - checks operator set registration
  - `operatorRegistered` - checks either registration type
  - `disableM2QuorumRegistration` - disables new M2 registrations
  - `updateOperatorSetsConfig` - updates operator set strategy parameters

### ECDSAAVSRegistrar
- Added AllocationManager integration:
  - `registerOperator` - handles operator registration
  - `deregisterOperator` - handles operator deregistration
  - `supportsAVS` - checks AVS support

### ECDSAServiceManagerBase
- Added IServiceManager implementations:
  - `deregisterOperatorFromOperatorSets`
  - `addPendingAdmin`
  - `removePendingAdmin`
  - `removeAdmin`
  - `setAppointee`
  - `removeAppointee`
- Added operator set support:
  - `getOperatorSetStrategies` - retrieves strategies for an operator set

### New Implementations
- Added `ECDSAStakeRegistryRatioWeight` with configurable weight ratios
- Updated `ECDSAStakeRegistryPermissioned` with new ejection logic

### Test Updates
- Updated existing test files:
  - `ECDSAStakeRegistryUnit.t.sol` 
  - `ECDSAStakeRegistryPermissionedUnit.t.sol` 
  - `ECDSAStakeRegistryEqualWeightUnit.t.sol` 
  - `ECDSAServiceManager.t.sol`

- Added new test file:
[...]

PR #440, opened 2025-03-18
feat: add operator sets to ecdsa registry and service manager

Sidu28:
**Motivation:**

**Modifications:**

**Result:**


jaxxjj:
I fixed the previous wrong logic and function for setting operator sets' strategy multipliers emitting the event `event OperatorSetStrategyParamsUpdated(uint32 indexed operatorSetId, StrategyParams[] params)` in [32c3097](https://github.com/Layr-Labs/eigenlayer-middleware/pull/411/commits/32c309742ff019bbe66cf96a395b421d0ff23179). I think a mapping `mapping(uint32 => mapping(address => uint256)) internal operatorSetStrategyMultipliers` and struct `struct OperatorSetStrategyMultiplier {
        uint32 operatorSetId;
        address strategy;
        uint256 multiplier;
    }` should set the multiplier for each strategy in operator sets and correctly apply it when calculating weights.

PR #445, opened 2025-03-24
perf: code size improvement for registry coordinator

stevennevins:
**Motivation:**

The registry coordinator is very close to the code size limit with standard compiler settings .  This leaves no room for small updates or compiler changes 

**Modifications:**

Updates the signature checker library and quorum bitmap history libs to use external function visibility

**Result:**

About 1.3kb in space recovered. This improves the devex for integrators


8sunyuan (approved):
LGTM, will be more expensive but valid tradeoff

PR #446, opened 2025-03-28
fix: storage layout regression

stevennevins:
**Motivation:**

regression in the src / rc storage layouts

**Modifications:**

Re-order the inheritance to preserve the storage layout from the v0.5.4 rewards release

**Result:**

Consistent storage layout between releases

### DIFF
--- a/src/RegistryCoordinator.sol
+++ b/src/RegistryCoordinator.sol
@@ -34,7 +34,7 @@ import {RegistryCoordinatorStorage} from "./RegistryCoordinatorStorage.sol";
  *
  * @author Layr Labs, Inc.
  */
-contract RegistryCoordinator is RegistryCoordinatorStorage, SlashingRegistryCoordinator {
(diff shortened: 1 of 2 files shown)


8sunyuan:
Compared storage inspects between `v1.1.1-testnet-slashing` and the current commit.

`SlashingRegistryCoordinator` looks good
image

8sunyuan:
But now the RegistryCoordinator storage layout is now appended at the end and is different from the previous `v1.1.1` release. Will this affect anyone?
image

stevennevins:
> But now the RegistryCoordinator storage layout is now appended at the end and is different from the previous `v1.1.1` release. Will this affect anyone? image

I don't know of anyone who upgraded on testnet, but we would need to create essentially a SlashingRegistryCoordinatorBase without EIP712 to resolve this, and then have RegistryCoordinator and SlashingRegistry coordinators inherit that separately to fully resolve this

8sunyuan (approved):
Noting that RegistryCoordinatorStorage has shifted down slots but this should be fine

PR #455, opened 2025-04-06
Feat: generate parameters for `checkSignatures` by referencing `OperatorStateRetriever` on-chain

rubydusa:
**Motivation:**

Computation of the parameters needed for signature verification is typically done through the Eigen go or rust services. However, the rust services are not reliable at the current moment and some projects might want to have signature verifications without having to rely on these services.

This PR aims to make the computation of the `NonSignerStakesAndSignature` parameter easier and accessible through implementing it in `OperatorStateRetriever` contract
 
**Modifications:**

- added `getNonSignerStakesAndSignature` to `OperatorStateRetriever` 
- added BN256G2 solidity library dependency

**Result:**

Computing the data necessary for performing signature verification is more accessible


src/OperatorStateRetriever.sol
```diff
@@ -231,4 +235,187 @@ contract OperatorStateRetriever {
             operators[i] = registryCoordinator.getOperatorFromId(operatorIds[i]);
         }
     }
+
+    // avoid stack too deep
+    struct GetNontSignerStakesAndSignatureMemory {
```
tsnewnami: nit; `GetNonSignerStakesAndSignatureMemory`?

src/OperatorStateRetriever.sol
```diff
@@ -231,4 +235,187 @@ contract OperatorStateRetriever {
...
+    ) external view returns (IBLSSignatureCheckerTypes.NonSignerStakesAndSignature memory) {
+        GetNontSignerStakesAndSignatureMemory memory m;
+        m.quorumApks = new BN254.G1Point[](quorumNumbers.length);
+        m.indexRegistry = registryCoordinator.indexRegistry();
+        m.blsApkRegistry = registryCoordinator.blsApkRegistry();
+
+        // Safe guard AVSs from generating NonSignerStakesAndSignature with invalid sigma
+        require(_isOnCurve(sigma), InvalidSigma());
+
+        m.signingOperatorIds = new bytes32[](operators.length);
```
tsnewnami: Comments could be useful for this loop here.

src/OperatorStateRetriever.sol
rubydusa: I assume by "nit" you mean "nitpick"?
I gave it this name because I think of this struct as an "extension" of the memory used for variables of the `getNonSignerStakesAndSignatute` - do you maybe have a suggestion for a better name for a struct used to avoid stack too deep?

tsnewnami: Right - makes sense. I didn't think the `t` was intentional. All good.

rubydusa: Oh lol didn't notice the T - ok got it

rubydusa: [2899e89](https://github.com/Layr-Labs/eigenlayer-middleware/pull/455/commits/2899e89fe236926cd4f314b1c10e40a7eb5a990c)

src/OperatorStateRetriever.sol
rubydusa: [235a7d1](https://github.com/Layr-Labs/eigenlayer-middleware/pull/455/commits/235a7d104e0eba2fb89b7b2bcff9a063a624b6cf)

src/OperatorStateRetriever.sol
```diff
@@ -27,6 +30,7 @@ contract OperatorStateRetriever {
     }
 
     error OperatorNotRegistered();
+    error InvalidSigma();
```
MegaRedHand:
```suggestion
    /// @dev Thrown when the signature is not on the curve.
    error InvalidSigma();
```

src/OperatorStateRetriever.sol
```diff
@@ -231,4 +235,188 @@ contract OperatorStateRetriever {
...
+        IIndexRegistry indexRegistry;
+        IBLSApkRegistry blsApkRegistry;
+        bytes32[] signingOperatorIds;
+    }
+
+    /**
+     * @notice Returns the stakes and signature information for non-signing operators in specified quorums
+     * @param registryCoordinator The registry coordinator contract to fetch operator information from
+     * @param quorumNumbers Array of quorum numbers to check for non-signers
+     * @param sigma The aggregate BLS signature to verify
```
MegaRedHand: We should mention this function can work even when the signature doesn't coincide with the given parameters.

test/unit/OperatorStateRetrieverUnit.t.sol
```diff
@@ -727,4 +729,600 @@ contract OperatorStateRetrieverUnitTests is MockAVSDeployer {
...
+        // Generate actual G2 pubkeys
+        BN254.G2Point memory op1G2 = _makeG2Point(2);
+        BN254.G2Point memory op2G2 = _makeG2Point(3);
+
+        // Mock the registry calls so the contract sees those G2 points
+        vm.mockCall(
+            address(blsApkRegistry),
+            abi.encodeWithSelector(IBLSApkRegistry.getOperatorPubkeyG2.selector, defaultOperator),
+            abi.encode(op1G2)
+        );
```
MegaRedHand: Why do we mock these? Can't we call them to get the actual values and use those?

test/unit/OperatorStateRetrieverUnit.t.sol
```diff
@@ -727,4 +729,600 @@ contract OperatorStateRetrieverUnitTests is MockAVSDeployer {
...
+        // Call under test
+        IBLSSignatureCheckerTypes.NonSignerStakesAndSignature memory result = operatorStateRetriever
+            .getNonSignerStakesAndSignature(
+            registryCoordinator, quorumNumbers, dummySigma, signingOperators, uint32(block.number)
+        );
+
+        // Validate
+        // One non-signer => otherOperator
+        assertEq(result.nonSignerQuorumBitmapIndices.length, 1, "Should have 1 non-signer");
+        assertEq(result.nonSignerPubkeys.length, 1, "Should have 1 non-signer pubkey");
```
MegaRedHand: Can we check the contents of these to ensure they're correct?

src/OperatorStateRetriever.sol
rubydusa: [ff9b6b9](https://github.com/Layr-Labs/eigenlayer-middleware/pull/455/commits/ff9b6b9c8875cbde61ca3fc4b221fb36b7d1d46a)

test/unit/OperatorStateRetrieverUnit.t.sol
rubydusa: Initiating the entire registration flow would be cumbersome and unnecessary for the scope of the tests. 
It's possible to extend the harness, but IMO harnesses introduce extra coupling with internal implementation details so it's preferable to mock instead

src/OperatorStateRetriever.sol
rubydusa: [dd8aaff](https://github.com/Layr-Labs/eigenlayer-middleware/pull/455/commits/dd8aaffb5c603ab7da3b9665a1bc8c99c2ecf849)

test/unit/OperatorStateRetrieverUnit.t.sol
rubydusa: [1e57364](https://github.com/Layr-Labs/eigenlayer-middleware/pull/455/commits/1e5736472ae23dd75b0460d4c5490c692a441e0d)

MegaRedHand:
CI's failing due to formatting: https://github.com/Layr-Labs/eigenlayer-middleware/actions/runs/14387747950/job/40347432364?pr=455

rubydusa:
There is invalid rpc url error in the CI which I can also see happens on latest commit of dev: https://github.com/Layr-Labs/eigenlayer-middleware/actions/runs/14312178100/job/40109659601#step:5:883

However this PR has 2 failing CI checks and the latest commit in dev only has 1, so I'm not sure if these are just simply different workflows or am I missing something

MegaRedHand:
The CI errors in this PR are due to external contributors not having access to repository secrets. I'm not sure about the failure in the latest `dev` commit, but it seems to be due to something similar. You can run the tests locally with:

```
HOLESKY_RPC_URL=<some-public-rpc> forge test -vvv
FOUNDRY_PROFILE=forktest forge test --match-contract Integration -vvv
```

src/libraries/BN256G2.sol
```diff
@@ -0,0 +1,339 @@
+// SPDX-License-Identifier: BUSL-1.1
```
ypatil12: Switch out for audited version

ypatil12 (commented):
First pass, minor comment

ypatil12 (dismissed):
Since the `OperatorStateRetriever` is already in audit, let's inherit from that and create an unaudited version with this code

src/libraries/BN256G2.sol
ypatil12: Synced offline, will keep in the `unaudited` folder

rubydusa:
Closing because of #461

PR #457, opened 2025-04-07
feat: OperatorStateRetrieverWithSocket

pakim249CAL:
**Motivation:**

Add getOperatorStateWithSocket entry points to negate the need for an additional query for the socket for offchain integrators.

This change was [merged ](https://github.com/Layr-Labs/eigenlayer-middleware/pull/418)to eigenda-dev. However, this should also probably be propagated over to here as well. That way, eigenda can update to a canonical branch in this repo.

**Modifications:**

### OperatorStateRetriever

* Added getOperatorStateWithSocket(
        ISlashingRegistryCoordinator registryCoordinator,
        bytes32 operatorId,
        uint32 blockNumber)
* Added getOperatorStateWithSocket(
        ISlashingRegistryCoordinator registryCoordinator,
        bytes memory quorumNumbers,
        uint32 blockNumber)

**Result:**

The OperatorStateRetriever will have these additional functions.

### DIFF
--- a/src/OperatorStateRetriever.sol
+++ b/src/OperatorStateRetriever.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.27;
 
 import {ISlashingRegistryCoordinator} from "./interfaces/ISlashingRegistryCoordinator.sol";
+import {ISocketRegistry} from "./interfaces/ISocketRegistry.sol";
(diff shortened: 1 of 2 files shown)


RonTuretzky:
bumping as this will make commonware intergation easier

PR #462, opened 2025-04-30
fix: ejection manager inclusive ejection

ypatil12:
**Motivation:**

Per Hexens Eigen2-2, The current rate limit logic allows ejection beyond the allowed limit.

**Modifications:**

Modify the ejection logic such that when the caller is the ejector, do not automatically eject the first operator regardless of limit. 

**Result:**

Proper rate limit assertions.


ypatil12:
Regression Test: 
Screenshot 2025-04-30 at 4 42 36 PM

pakim249CAL (approved):

wadealexc (approved):
LGTM

PR #463, opened 2025-04-30
fix: lookahead period constraint

ypatil12:
**Motivation:**

The `StakeRegistry` is missing a constraint check for modification of lookahead time
of slashable quorum. This results in the Lookahead period potentially being invalid. 

**Modifications:**

Update `_setLookAheadPeriod` to revert if the `_lookaheadBlocks` is greater than the `DEALLOCATION_DELAY`. Note that we'll also have to update this check in `_createQuorum` since it does a `>=` check. Operators are slashable for the entire delay, inclusive of the last block. 

**Result:**

Proper constraints.

### DIFF
--- a/src/StakeRegistry.sol
+++ b/src/StakeRegistry.sol
@@ -798,7 +798,7 @@ contract StakeRegistry is StakeRegistryStorage {
             QuorumNotSlashable()
         );
         require(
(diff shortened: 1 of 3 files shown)


pakim249CAL (approved):

wadealexc (approved):
LGTM

PR #464, opened 2025-05-01
fix: enforce quorum registration on churn

ypatil12:
**Motivation:**

It is possible to break the `maxOperatorCount` invariant by doing the following:

Let's assume there are two quorums, 1 and 2, with a `maxOperatorCount` of 2. 

1. Alice & Bob register for quorum 1
2. Bob registers for quorum 2
3. Bob deregisters from quorum 1, Charlie enters
4. Quorum 1 Members: Alice/Charlie. Quorum 2 members: Bob
5. Eve creates a churn registration that exits Bob.  Quorum 1 has 3 members. This works just fine since we allow a churn to occur if the `operatorToKick` is registered to the AVS (not the quorum): https://github.com/Layr-Labs/eigenlayer-middleware/blob/f5adbcac55d9336cd646ce71bc467aa7e20f1a12/src/SlashingRegistryCoordinator.sol#L405-L408

Although this assumes that the `churnApprover` is buggy, we should still be enforcing that you are churning a user if they are registered for the quorum. 

**Modifications:**

Require that the `operatorToKick` is registered for the quorum. 

**Result:**

Stricter churn guarantees.

### DIFF
--- a/src/SlashingRegistryCoordinator.sol
+++ b/src/SlashingRegistryCoordinator.sol
@@ -705,8 +705,11 @@ contract SlashingRegistryCoordinator is
         bytes32 idToKick = _operatorInfo[operatorToKick].operatorId;
         require(newOperator != operatorToKick, CannotChurnSelf());
         require(kickParams.quorumNumber == quorumNumber, QuorumOperatorCountMismatch());
+
+        uint192 quorumBitmap;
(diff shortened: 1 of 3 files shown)


wadealexc (approved):
LGTM

PR #474, opened 2025-05-07
fix: remove stale stakes flag

ypatil12:
**Motivation:**

The stale stakes flag is buggy since it doesn't properly get the `quorumUpdateBlockNumber` in the given range. The fix is to add checkpointed storage for the latest quorum update block number. Since we are moving towards a cert-verifier world, it's simpler to remove this feature rather than ship checkpointed storage OR broken code. 

**Modifications:**

Removed `staleStakesForbidden`. 

**Result:**

Cleaner code


wadealexc (approved):

ypatil12:
Note: Storage checker failing due to changing the naming of `staleStakesForbidden` to be deprecated. This is OK

PR #479, opened 2025-05-16
refactor: singleton cv combining ECDSA and BN254

8sunyuan:
**Motivation:**

Redesigning to have a singleton CertificateVerifier that is deployed per destination chain where the AVSs utilizing this contract trust the `globalOperatorTableRoot` is valid and the EigenDA operatorSet which configures this in a timely basis.

**Modifications:**

- Combined the ECDSACertificateVerifier and the BN254CertificateVerifier interfaces into CertificateVerifier.
- OperatorSet configuration parameters are now  configured on the L1 via a `CrossChainRegistry` contract and transported by the following functions.
    - `ICertificateVerifier.updateECDSAOperatorTable` or `ICertificateVerifier.updateBN254OperatorTable`

**Result:**

Standarizing interface with a canonical CertVerifier contract and less deployment duplication from AVSs.


eigenmikem (approved):
Looks great

src/interfaces/multichain/ICertificateVerifier.sol
```diff
@@ -15,46 +98,160 @@ interface ICertificateVerifierErrors {
...
+     * @return whether or not certificate is valid and meets thresholds
+     */
+    function verifyECDSACertificateNominal(
+        ECDSACertificate memory cert,
+        uint96[] memory totalStakeNominalThresholds
+    ) external returns (bool);
+
+    /* BN254 CERTIFICATE VERIFIER INTERFACE */
+
+    /**
```
ypatil12, line 200: No ejection?

ypatil12 (changes requested):
Thoughts on separating out the interfaces, and then combining into one global `ICertificateVerifier`

src/interfaces/multichain/ICertificateVerifier.sol
8sunyuan: https://github.com/Layr-Labs/eigenlayer-middleware/pull/479/commits/93aba5e47a5c686d3fb89b93e2068fba3a1a6d73

8sunyuan:
> Thoughts on separating out the interfaces, and then combining into one global `ICertificateVerifier`

Imo, less confusing if its one interface for a single contract but not too opposed

ypatil12 (approved):

PR #480, opened 2025-05-16
fix: bls sig operator state retriever sorting

RonTuretzky:
**Motivation:**

Non-signer pubkeys were not properly ordered in the last iteration of this feature

**Modifications:**

Added OZ libraries as packed utilities for sorting, added sorting of the non-signers before returning them 

**Result:**
Now all parameters returned are sorted in ascending order, which allows for strong assurances regarding non-duplicates in the array


src/unaudited/BLSSigCheckUtils.sol
```diff
@@ -0,0 +1,801 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.20;
+
+import {BN254} from "../libraries/BN254.sol";
+import {StorageSlot} from "@openzeppelin/contracts/utils/StorageSlot.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+/**
```
ypatil12, line 12: Nit: can we add a comment why we're copy-pasting these (due to not being on v5.1 OZ). Another option is to add another submodule

ypatil12 (commented):
Minor comment, will approve after

src/unaudited/BLSSigCheckUtils.sol
RonTuretzky: addressed with 85ed41d

RonTuretzky:
image
All tests are passing locally (including testnet)

ypatil12 (approved):
LGTM. Will fix CI separately. Tests pass locally. This is an issue with fork PRs

PR #481, opened 2025-05-19
feat: keyregistrar

eigenmikem:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/KeyRegistrar.sol
+++ b/src/KeyRegistrar.sol
@@ -1,0 +1,472 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.27;
+
+import {Initializable} from "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
+import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+import {BN254} from "./libraries/BN254.sol";
+
+/**
+ * @title KeyRegistrar
+ * @notice A core singleton contract that manages operator keys for different AVSs
+ * @dev Provides registration, deregistration, and rotation of keys with support for aggregate keys
+ */
+contract KeyRegistrar is Initializable, OwnableUpgradeable {
+    using BN254 for BN254.G1Point;
+
+    /// @dev Returns the hash of the zero pubkey for BN254 aka BN254.G1Point(0,0)
+    bytes32 internal constant ZERO_PK_HASH = hex"ad3228b676f7d3cd4284a5443f17f1962b36e491b30a40b2405849e597ba5fb5";
+
+    /// @dev Enum defining supported curve types
+    enum CurveType {
+        ECDSA,
+        BN254
+        // BLS12_381 // Future support
+    }
+
+    /// @dev Structure to store key information across different curves
+    struct KeyInfo {
+        bool isRegistered;
+        uint256 lastRotationBlock;
+    }
+
+    /// @dev Maps (AVS, operator, curve type) to key info
+    mapping(address => mapping(address => mapping(CurveType => KeyInfo))) private avsOperatorKeyInfo;
+
+    /// @dev Maps (AVS, operator) to their ECDSA public keys
+    mapping(address => mapping(address => bytes)) private avsOperatorToECDSAKey;
+
+    /// @dev Maps (AVS, operator) to their BN254 G1 points
+    mapping(address => mapping(address => BN254.G1Point)) private avsOperatorToBN254Key;
+
+    /// @dev Maps (AVS, operator) to their BN254 G2 points (for BLS verification)
+    mapping(address => mapping(address => BN254.G2Point)) private avsOperatorToBN254KeyG2;
+
+    /// @dev Maps (AVS, operator, curve type) to key hash for quick lookup
+    mapping(address => mapping(address => mapping(CurveType => bytes32))) private avsOperatorToKeyHash;
+
+    /// @dev Maps (AVS, curve type, key hash) to operator address for reverse lookup
+    mapping(address => mapping(CurveType => mapping(bytes32 => address))) private avsKeyHashToOperator;
+
+    /// @dev Maps (AVS, operatorSetId) to their aggregate BN254 G1 point
+    mapping(address => mapping(uint32 => BN254.G1Point)) private avsOperatorSetToAggregateBN254Key;
+
+    /// @dev Maps AVS to their authorized registrar contract
+    mapping(address => address) public avsToRegistrar;
+
+    /// @dev Optional delay period before a key rotation takes effect (in blocks)
+    uint256 public rotationDelay;
+
+    /// Events
+    event RegistrarAuthorized(address indexed avs, address indexed registrar);
+    event KeyRegistered(address indexed avs, address indexed operator, CurveType curveType, bytes pubkey);
+    event KeyDeregistered(address indexed avs, address indexed operator, CurveType curveType);
+    event KeyRotationInitiated(address indexed avs, address indexed operator, CurveType curveType, bytes oldKey, bytes newKey, uint256 effectiveBlock);
+    event AggregateBN254KeyUpdated(address indexed avs, uint32 indexed operatorSetId, BN254.G1Point newAggregateKey);
+    event RotationDelayUpdated(uint256 oldDelay, uint256 newDelay);
+
+    /// Errors
+    error KeyAlreadyRegistered();
+    error KeyNotRegistered();
+    error InvalidKeyFormat();
+    error ZeroAddress();
+    error ZeroPubkey();
+    error KeyRotationInProgress();
+    error InvalidCurveType();
+    error Unauthorized();
+    error OperatorNotFound();
+
+    /// @dev Only the authorized registrar for an AVS can call functions with this modifier
+    modifier onlyAVSRegistrar(address avs) {
+        if (msg.sender != avsToRegistrar[avs] && msg.sender != owner()) {
+            revert Unauthorized();
+        }
+        _;
+    }
+
+    /// @dev Initializes the contract
+    /// @param initialRotationDelay Initial key rotation delay in blocks
+    function initialize(uint256 initialRotationDelay) external initializer {
+        __Ownable_init();
+        rotationDelay = initialRotationDelay;
+    }
+
+    /**
+     * @notice Authorize a registrar contract for a specific AVS
+     * @param registrar Address of the registrar contract
+     */
+    function authorizeRegistrar(address registrar) external {
+        if (registrar == address(0)) {
+            revert ZeroAddress();
+        }
+        avsToRegistrar[msg.sender] = registrar;
+        emit RegistrarAuthorized(msg.sender, registrar);
+    }
+
+    /**
+     * @notice Registers a new cryptographic key for an operator with a specific AVS
+     * @param avs Address of the AVS
+     * @param operator Address of the operator
+     * @param operatorSetIds Array of operator set IDs to add the operator to (only for BN254 keys)
+     * @param curveType Type of curve (ECDSA, BN254)
+     * @param pubkey Public key bytes
+     * @return alreadyRegistered True if the key was already registered
+     */
+    function registerKey(
+        address avs,
+        address operator,
+        uint32[] calldata operatorSetIds,
+        CurveType curveType,
+        bytes calldata pubkey
+    ) external onlyAVSRegistrar(avs) returns (bool alreadyRegistered) {
+        if (avs == address(0) || operator == address(0)) {
+            revert ZeroAddress();
+        }
+        
+        // Check if the key is already registered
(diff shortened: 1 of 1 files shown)


src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    /// @dev Structure to store key information across different curves
+    struct KeyInfo {
+        bool isRegistered;
+        uint256 lastRotationBlock;
+    }
+
+    /// @dev Maps (AVS, operator, curve type) to key info
+    mapping(address => mapping(address => mapping(CurveType => KeyInfo))) private avsOperatorKeyInfo;
```
ypatil12: Let's change to mapping from operatorSet to KeyInfo, then we can get rid of `CurveType` in this mapping

ypatil12: (Assume single key type per operatorSet)

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+    mapping(address => mapping(address => bytes)) private avsOperatorToECDSAKey;
+
+    /// @dev Maps (AVS, operator) to their BN254 G1 points
+    mapping(address => mapping(address => BN254.G1Point)) private avsOperatorToBN254Key;
+
+    /// @dev Maps (AVS, operator) to their BN254 G2 points (for BLS verification)
+    mapping(address => mapping(address => BN254.G2Point)) private avsOperatorToBN254KeyG2;
+
+    /// @dev Maps (AVS, operator, curve type) to key hash for quick lookup
+    mapping(address => mapping(address => mapping(CurveType => bytes32))) private avsOperatorToKeyHash;
```
ypatil12: TODO: identify need for using key hash

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+
+    /// @dev Maps (AVS, curve type, key hash) to operator address for reverse lookup
+    mapping(address => mapping(CurveType => mapping(bytes32 => address))) private avsKeyHashToOperator;
+
+    /// @dev Maps (AVS, operatorSetId) to their aggregate BN254 G1 point
+    mapping(address => mapping(uint32 => BN254.G1Point)) private avsOperatorSetToAggregateBN254Key;
+
+    /// @dev Maps AVS to their authorized registrar contract
+    mapping(address => address) public avsToRegistrar;
```
ypatil12: Let's remove this and just use `PermissionController`

src/KeyRegistrar.sol
RonTuretzky: why not make KeyInfo bytes so it can support any curve type in the future without upgrading this contract? 
Same arg could be made to use an int/string for the CurveType key , allows AVS teams to extend this functionality without EL intervention

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+    /**
+     * @notice Registers a new cryptographic key for an operator with a specific AVS
+     * @param avs Address of the AVS
+     * @param operator Address of the operator
+     * @param operatorSetIds Array of operator set IDs to add the operator to (only for BN254 keys)
+     * @param curveType Type of curve (ECDSA, BN254)
+     * @param pubkey Public key bytes
+     * @return alreadyRegistered True if the key was already registered
+     */
+    function registerKey(
```
ypatil12, line 131: If operators register their key first, I'm wondering if we can then just have a separate ECDSA and BN254 functions? Unsure what is cleaner...

ypatil12 (commented):
First pass

src/KeyRegistrar.sol
RonTuretzky: This would also allow support for verifiable DKG in the future

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+    mapping(address => mapping(address => BN254.G2Point)) private avsOperatorToBN254KeyG2;
+
+    /// @dev Maps (AVS, operator, curve type) to key hash for quick lookup
+    mapping(address => mapping(address => mapping(CurveType => bytes32))) private avsOperatorToKeyHash;
+
+    /// @dev Maps (AVS, curve type, key hash) to operator address for reverse lookup
+    mapping(address => mapping(CurveType => mapping(bytes32 => address))) private avsKeyHashToOperator;
+
+    /// @dev Maps (AVS, operatorSetId) to their aggregate BN254 G1 point
+    mapping(address => mapping(uint32 => BN254.G1Point)) private avsOperatorSetToAggregateBN254Key;
```
RonTuretzky, line 47: As this is specific to BLS, If we adopt the extensible pattern I suggested above, this would have to be in a "additional data" field or metadata field

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+
+    /// @dev Maps AVS to their authorized registrar contract
+    mapping(address => address) public avsToRegistrar;
+
+    /// @dev Optional delay period before a key rotation takes effect (in blocks)
+    uint256 public rotationDelay;
```
RonTuretzky: What is the benefit of introducing a rotation delay? I fear that in 0 day scenarios , allowing the operator key to override the signing key with no delay might have some edge case advantage

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+            }
+        } else {
+            revert InvalidCurveType();
+        }
+
+        // Update rotation timestamp
+        uint256 effectiveBlock = block.number + rotationDelay;
+        avsOperatorKeyInfo[avs][operator][curveType].lastRotationBlock = effectiveBlock;
+
+        emit KeyRotationInitiated(avs, operator, curveType, oldKey, newPubkey, effectiveBlock);
```
RonTuretzky: the way this could be handled in the extensible pattern is when someone registers a new curve type in the keyregistrar (permissionlessly) they also point to a handling function which then can decode/encode/validate the data as necessary and implements an interface. 

Another option is to let the AVS handle this and simple not include any encoding/decoding logic in this contract, just storing arbitrary bytes. This is an anti-pattern to the desired behavior of not having AVS teams re-deploy the same code though. 

I prefer the first approach I suggested, as it would still allow Eigen to support first class curves without the tradeoff of limiting extensibility

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+        if (isAddition) {
+            newApk = currentApk.plus(key);
+        } else {
+            newApk = currentApk.plus(key.negate());
+        }
+        
+        avsOperatorSetToAggregateBN254Key[avs][operatorSetId] = newApk;
+        
+        emit AggregateBN254KeyUpdated(avs, operatorSetId, newApk);
+    }
```
RonTuretzky: This would be code that would sit outside this contract in the pattern I suggested above , maintained by eigen, and when someone registers a "Key Type" enum/identifier , they would be able to point to an external contract with this handling logic (that would receive a bytes array and decode it to BN254.G1Point, in this case)

src/KeyRegistrar.sol
RonTuretzky: yea I think there needs to be one register function that receives bytes and a curve type, and does a look up for the dedicated handler of that reg type

src/KeyRegistrar.sol
gpsanant, line 47: why store the aggregate in the contract instead of calculating it in the calculator?

src/KeyRegistrar.sol
gpsanant: why not just reuse registrar from AM?

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+
+    /**
+     * @notice Registers a BN254 key for an operator with a specific AVS
+     * @param avs Address of the AVS
+     * @param operator Address of the operator
+     * @param pubkey BN254.G1Point and BN254.G2Point encoded as bytes
+     * @return The registered BN254.G1Point
+     */
+    function _registerBN254Key(
```
gpsanant: they need to provide a signature here as well to prevent https://xn--2-umb.com/22/bls-signatures/#rogue-key-attack

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,472 @@
...
+        return avsOperatorToKeyHash[avs][operator][curveType];
+    }
+
+    /**
+     * @notice Registers an ECDSA key for an operator with a specific AVS
+     * @param avs Address of the AVS
+     * @param operator Address of the operator
+     * @param pubkey ECDSA public key bytes
+     */
+    function _registerECDSAKey(
```
gpsanant: why isn't the pubkey decoded to address?

src/KeyRegistrar.sol
gpsanant: unclear why the core needs to enforce a rotation delay? any AVS that cares about a rotation delay knows enough to implement it themselves?

src/KeyRegistrar.sol
gpsanant: yeah im unclear on why we need this too

src/KeyRegistrar.sol
gpsanant, line 131: can we have the operator call these functions directly and have the AVS just view from this contract? reduces middleware scope of concern and codebase size...

src/KeyRegistrar.sol
RonTuretzky: this would introduce an "updateOperators" like pattern for maintaining the apk... imo better to do it in here , but keep it in an `extraBytes` field

src/KeyRegistrar.sol
ypatil12: This was a feature requested by AVS devs

src/KeyRegistrar.sol
ypatil12: 2nd draft is going to do this :)

src/KeyRegistrar.sol
RonTuretzky: Key rotation is very useful for TEE security best practices , key compromises, machine failures etc

src/KeyRegistrar.sol
eigenmikem: Yea that's going in the 2nd draft, I can port over most of the code from the certVerifier

src/KeyRegistrar.sol
eigenmikem: Was requested in the TDD

src/KeyRegistrar.sol
gpsanant: key rotation makes sense, but it's not clear to me why a rotation delay needs to be enforced by the core protocol?

src/KeyRegistrar.sol
RonTuretzky: > Was requested in the TDD

I am highly opposed to introducing another `updateOperators` pattern as that one has been fairly unadopted and has cost hundreds of thousands of dollars cc @ypatil12 @gpsanant

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,717 @@
...
+            if (newPubkey[i] != 0) {
+                isZero = false;
+                break;
+            }
+        }
+        if (isZero) revert ZeroPubkey();
+        
+        // Check global uniqueness for new key
+        bytes32 newKeyHash = keccak256(newPubkey);
+        if (globalKeyRegistry[newKeyHash]) {
```
8sunyuan, line 403: nit: replace `revert Error()` with just a single require statement

Also noticing multiple code blocks where you have the following code:
```
if (globalKeyRegistry[newKeyHash]) {
    revert KeyAlreadyRegistered();
}
globalKeyRegistry[newKeyHash] = true;
```
Maybe just place this in a `_setKeyHash(bytes32 keyhash)` internal func

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    /**
+     * @notice Checks if a key is registered for an operator with a specific AVS
+     * @param avs Address of the AVS
+     * @param operatorSetId ID of the operator set
+     * @param operator Address of the operator
+     * @return True if the key is registered
+     */
+    function isRegistered(
```
8sunyuan, line 503: nit: rename `isKeyRegistered` for clarity?

src/KeyRegistrar.sol
```diff
@@ -0,0 +1,717 @@
...
+        if (newPubkey[0] != 0x04) revert InvalidKeyFormat();
+        
+        bool isZero = true;
+        for (uint i = 1; i < 65; i++) {
+            if (newPubkey[i] != 0) {
+                isZero = false;
+                break;
+            }
+        }
+        if (isZero) revert ZeroPubkey();
```
8sunyuan, line 399: could we just have some internal constant storing the hash of bytes pubkey[1:] = [0x00, ..., 0x00]
and compare if keccak256(abi.encode(pubkey)) is equal to that? Would be a lot more efficient vs this check

ypatil12:
Closing PR, as this has moved to Core

PR #490, opened 2025-07-01
chore: remove unused imports

ypatil12:
**Motivation:**

The `AVSRegistrarAsIdentifier` had unnecessary imports

**Modifications:**

- Remove `Initializable` and `SocketRegistry`

**Result:**

Cleaner code


8sunyuan (approved):
lgtm, storage layout seems to be as expected since the socket storage shouldn't have been there in first place

PR #493, opened 2025-07-09
fix: table calc interface

ypatil12:
**Motivation:**

Update the interface in the table calculator for `getOperatorSetWeights` and `getOperatorWeights`
**Modifications:**

- `getOperatorWeights` -> `getOperatorSetWeights`
- `getOperatorWeight` -> `getOperatorWeights`, and now returns an array of weights
- Bump on core version and fix integration test setup

**Result:**

Correct code


nadir-akhtar (approved):
Seems the storage layout checker is complaining given that the SlashEscrowFactory had been removed, but LGTM

PR #504, opened 2025-07-18
fix(H-1): correct array indexing for BN254TableCalculatorBase._calculateOperatorTable

nadir-akhtar:
**Motivation:**

As discovered during audit, `BN254TableCalculatorBase._calculateOperatorTable` incorrectly uses `i` for the index when collecting registered operators rather than `operatorCount`, leading to malformed arrays. The corresponding test intended to validate this behavior but failed to.

**Modifications:**

* Correct contract implementation
* Fix test to correctly fail when array is malformed, and pass when it is correctly constructed and the root matchees

**Result:**

Auditing finding addressed, correct code and tests


ypatil12 (approved):
LGTM

PR #509, opened 2025-07-23
fix(L-1): improve `updateSocket` check

0xClandestine:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/middlewareV2/registrar/modules/SocketRegistry.sol
+++ b/src/middlewareV2/registrar/modules/SocketRegistry.sol
@@ -4,6 +4,10 @@ pragma solidity ^0.8.27;
 import {ISocketRegistry} from "../../../interfaces/ISocketRegistryV2.sol";
 import {SocketRegistryStorage} from "./SocketRegistryStorage.sol";
 import {
+    ISlashingRegistryCoordinator,
+    ISlashingRegistryCoordinatorTypes
+} from "../../../interfaces/ISlashingRegistryCoordinator.sol";
+import {
     OperatorSetLib,
     OperatorSet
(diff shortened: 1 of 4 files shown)


src/middlewareV2/registrar/modules/SocketRegistry.sol
```diff
@@ -22,7 +30,11 @@ abstract contract SocketRegistry is SocketRegistryStorage {
 
     /// @inheritdoc ISocketRegistry
     function updateSocket(address operator, string memory socket) external {
-        require(msg.sender == operator, CallerNotOperator());
+        require(
+            slashingRegistryCoordinator.getOperatorStatus(operator)
+                == ISlashingRegistryCoordinatorTypes.OperatorStatus.REGISTERED,
+            CallerNotOperator()
+        );
```
0xClandestine, line 37: Figured using slashRegistryCoordinator made more sense than ALM, can change if needed.

ypatil12: Doesn't make sense here, because the AVS may not deploy the SlashingRegistryCoordiantor. It's assumed that all the storage for operator set membership lives in the `AllocationManager`

Now that I think about it, there's no way to validate that an operator is part of an AVS, only the operatorSet...

0xClandestine: Oh assumed all V2 related stuff inherently had slashing. Curious how we want to address this.

ypatil12: Given that there is no way to introspect allocationManager membership, no way for us to address this

0xClandestine:
> Given that there is no way to introspect allocationManager membership, no way for us to address this

PR #512, opened 2025-07-23
fix(I-04): make avs var stateful & remove confusing natspec 

0xClandestine:
**Motivation:**

The following natspec is present but isn't true and suggest improper use.

```solidity
/// @dev The immutable avs address `AVSRegistrar` is NOT the address of the AVS in EigenLayer core.
/// @dev The address of the AVS in EigenLayer core is the proxy contract, and it is set via the `initialize` function below.
```

**Modifications:**

- Removed confusing comments
- Made AVS var stateful and `AVSRegistrar` abstract

**Result:**

Improved clarity.

### DIFF
--- a/src/middlewareV2/registrar/AVSRegistrar.sol
+++ b/src/middlewareV2/registrar/AVSRegistrar.sol
@@ -31,7 +31,7 @@ abstract contract AVSRegistrar is Initializable, AVSRegistrarStorage {
         _disableInitializers();
     }
 
-    // @dev This MUST be added to an `initialize` function in child contracts.
+    /// @dev This initialization function MUST be added to a child's `initialize` function to avoid uninitialized storage.
     function __AVSRegistrar_init(
         address _avs
     ) internal virtual {
--- a/src/middlewareV2/registrar/AVSRegistrar.sol
+++ b/src/middlewareV2/registrar/AVSRegistrar.sol
@@ -25,11 +25,17 @@ contract AVSRegistrar is Initializable, AVSRegistrarStorage {
     }
 
     constructor(
-        address _avs,
         IAllocationManager _allocationManager,
         IKeyRegistrar _keyRegistrar
-    ) AVSRegistrarStorage(_avs, _allocationManager, _keyRegistrar) {
+    ) AVSRegistrarStorage(_allocationManager, _keyRegistrar) {
         _disableInitializers();
+    }
+
+    // @dev This MUST be added to an `initialize` function in child contracts.
+    function __AVSRegistrar_init(
+        address _avs
+    ) internal virtual {
+        avs = _avs;
     }
 
     /// @inheritdoc IAVSRegistrar
@@ -67,11 +73,6 @@ contract AVSRegistrar is Initializable, AVSRegistrarStorage {
         address _avs
     ) public view virtual returns (bool) {
         return _avs == avs;
-    }
-
-    /// @inheritdoc IAVSRegistrarInternal
-    function getAVS() external view virtual returns (address) {
-        return avs;
     }
 
     /*
--- a/src/middlewareV2/registrar/AVSRegistrar.sol
+++ b/src/middlewareV2/registrar/AVSRegistrar.sol
@@ -16,7 +16,7 @@ import {AVSRegistrarStorage} from "./AVSRegistrarStorage.sol";
 import {Initializable} from "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 
 /// @notice A minimal AVSRegistrar contract that is used to register/deregister operators for an AVS
-contract AVSRegistrar is Initializable, AVSRegistrarStorage {
+abstract contract AVSRegistrar is Initializable, AVSRegistrarStorage {
     using OperatorSetLib for OperatorSet;
 
     modifier onlyAllocationManager() {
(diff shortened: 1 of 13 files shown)


src/middlewareV2/registrar/presets/AVSRegistrarAsIdentifier.sol
```diff
@@ -19,8 +19,6 @@ contract AVSRegistrarAsIdentifier is AVSRegistrar {
     /// @notice The permission controller for the AVS
     IPermissionController public immutable permissionController;
 
-    /// @dev The immutable avs address `AVSRegistrar` is NOT the address of the AVS in EigenLayer core.
```
ypatil12, line 20: This fix here shouldn't be to remove the comment. It should be:

> Consider enforcing upon initialization that avs==address(this) to avoid potential issues.

ypatil12 (changes requested):

src/middlewareV2/registrar/presets/AVSRegistrarAsIdentifier.sol
```diff
@@ -50,9 +51,4 @@ contract AVSRegistrarAsIdentifier is AVSRegistrar {
     ) public view override returns (bool) {
```
ypatil12: Don't need to override this

test/unit/middlewareV2/AllowlistUnit.t.sol
```diff
@@ -1,479 +1,479 @@
...
-
-        // Verify operator2 is not in the array
-        for (uint256 i = 0; i < allowedOperators.length; i++) {
-            assertTrue(
-                allowedOperators[i] != operator2, "Removed operator should not be in the array"
-            );
-        }
-    }
-}
+// // SPDX-License-Identifier: BUSL-1.1
```
nadir-akhtar: Why did we comment the whole file out?

src/middlewareV2/registrar/AVSRegistrar.sol
```diff
@@ -28,13 +28,19 @@ contract AVSRegistrar is Initializable, AVSRegistrarStorage {
...
     constructor(
-        address _avs,
         IAllocationManager _allocationManager,
         IKeyRegistrar _keyRegistrar
-    ) AVSRegistrarStorage(_avs, _allocationManager, _keyRegistrar) {
+    ) AVSRegistrarStorage(_allocationManager, _keyRegistrar) {
         _disableInitializers();
     }
 
+    // @dev This MUST be added to an `initialize` function in child contracts.
```
nadir-akhtar: @0xClandestine Can we specify exactly why this must be? Something like:
```suggestion
    // @dev This MUST be added to an `initialize` function to avoid circular deployment dependency between this implementation and its corresponding ERC-1967 proxy.
```

0xClandestine: Not sure I understand your comment, AVSRegistrar is inherited by various presents who need varying parameters in their initialize function, thus initialize cannot be implemented in the AVSRegistrar. As an alternative I can update the AVSRegistrar to be abstract so this is better communicated.

test/unit/middlewareV2/AllowlistUnit.t.sol
0xClandestine: [8c3a508](https://github.com/Layr-Labs/eigenlayer-middleware/pull/512/commits/8c3a508bbae2cd62a25356df00ae03eb59025f97)

src/middlewareV2/registrar/presets/AVSRegistrarAsIdentifier.sol
0xClandestine: [ccbe836](https://github.com/Layr-Labs/eigenlayer-middleware/pull/512/commits/ccbe8369ae1802d3dd17bdee88c44e03e12179cb)

src/middlewareV2/registrar/presets/AVSRegistrarAsIdentifier.sol
0xClandestine: [5f664b1](https://github.com/Layr-Labs/eigenlayer-middleware/pull/512/commits/5f664b11a1251e1998e42b290cfe3f0f79123228)

src/middlewareV2/registrar/AVSRegistrar.sol
0xClandestine: [2aab0c0](https://github.com/Layr-Labs/eigenlayer-middleware/pull/512/commits/2aab0c0008b183a861955751d89e34df5e250f97)

ypatil12: AVS Registrar is abstract though?

nadir-akhtar: Ah sorry, I misinterpreted the intent of the comment @0xClandestine -- I thought it was why an initialization function was merited in the first place (as opposed to stating that the function needs to be inherited and implemented). An updated suggestion for clarity:

```suggestion
    // @dev This initialization function MUST be added to a child's `initialize` function to avoid uninitialized storage.
```

0xClandestine: [d33fd9d](https://github.com/Layr-Labs/eigenlayer-middleware/pull/512/commits/d33fd9d714a6b2bdd27d0a4b93080e401141f023)

0xClandestine:
storage test failing because avs is now mutable

ypatil12 (approved):

PR #521, opened 2025-07-31
docs: natspec updates

eigenmikem:
**Motivation:**

**Modifications:**

**Result:**

### DIFF
--- a/src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
+++ b/src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
@@ -41,9 +41,6 @@ abstract contract BN254TableCalculatorBase is IBN254TableCalculator {
     }
 
     /// @inheritdoc IOperatorTableCalculator
-    /**
-     * @dev Returns ABI-encoded BN254OperatorSetInfo for cross-chain compatibility
-     */
     function calculateOperatorTableBytes(
         OperatorSet calldata operatorSet
     ) external view virtual returns (bytes memory operatorTableBytes) {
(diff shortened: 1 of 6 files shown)


src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
```diff
@@ -37,20 +41,29 @@ abstract contract BN254TableCalculatorBase is IBN254TableCalculator {
...
+     */
     function calculateOperatorTableBytes(
         OperatorSet calldata operatorSet
     ) external view virtual returns (bytes memory operatorTableBytes) {
         return abi.encode(_calculateOperatorTable(operatorSet));
     }
 
     /// @inheritdoc IOperatorTableCalculator
+    /**
+     * @dev Returns operator addresses and their corresponding weight arrays for the operatorSet
```
ypatil12: I think the return statement in the `IOperatorTableCalculator` is sufficient for all of these? NO need for the dev comment?

src/middlewareV2/tableCalculator/ECDSATableCalculatorBase.sol
```diff
@@ -21,34 +21,50 @@ abstract contract ECDSATableCalculatorBase is IECDSATableCalculator {
...
+     */
     function getOperatorSetWeights(
         OperatorSet calldata operatorSet
     ) external view virtual returns (address[] memory operators, uint256[][] memory weights) {
         return _getOperatorWeights(operatorSet);
     }
 
     /// @inheritdoc IOperatorTableCalculator
+    /**
+     * @dev Returns the weight array for a specific operator in the operatorSet, or empty array if not found
```
ypatil12: Think we can remove all of these return dev statements from the `IOperatorTableCalculator`

ypatil12 (commented):
2 comments

ypatil12 (approved):

PR #523, opened 2025-08-01
fix: correct gap from 48 -> 49

nadir-akhtar:
**Motivation:**

Gap is set to 48, but with only 1 storage variable, the gap should be 49 (50 - 1).

**Modifications:**

* Update gap from 48 to 49

**Result:**

Correct gap calculation


nadir-akhtar:
The dynamic array takes up 32 bytes, meaning the struct takes up two separate storage slots, meaning 48 is correct

PR #532, opened 2025-08-19
feat: add UAM to `SocketRegistry`

ypatil12:
**Motivation:**

In the `SocketRegistry`, only the operator can update the socket, and not an admin or appointee.

**Modifications:**

- Inherit `PermissionControllerMixin` in `SocketRegistry`
- Update constructor params for `TaskAVSRegistrarBase` and `AVSRegistrarWithSocket`

**Result:**

Clearer code

### DIFF
--- a/docs/middlewareV2/AVSRegistrar.md
+++ b/docs/middlewareV2/AVSRegistrar.md
@@ -302,9 +302,12 @@ function getOperatorSocket(
 
 ```solidity
 /**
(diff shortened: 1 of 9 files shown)


0xrajath (approved):

nadir-akhtar (approved):
Generally, let's flag these added permissions to users (perhaps in release notes), in case they want to rescope their existing admin keys based on these new privileges

PR #533, opened 2025-08-23
fix: add allowlist to TaskAVSRegistarBase

nadir-akhtar:
**Motivation:**

To ensure that AVSs have control over the aggregators registering for Hourglass, an allowlist is added to prevent arbitrary aggregator registration.

**Modifications:**

Inherited `Allowlist` into the TaskAVSRegistarBase

**Result:**

Stronger permissions around registration

### DIFF
--- a/src/avs/task/TaskAVSRegistrarBase.sol
+++ b/src/avs/task/TaskAVSRegistrarBase.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
 import {Initializable} from "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import {IAllocationManager} from
     "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
@@ -22,7 +21,6 @@ import {OperatorSet} from "eigenlayer-contracts/src/contracts/libraries/Operator
  */
 abstract contract TaskAVSRegistrarBase is
     Initializable,
-    OwnableUpgradeable,
     AVSRegistrarWithSocket,
     Allowlist,
     TaskAVSRegistrarBaseStorage
@@ -53,8 +51,6 @@ abstract contract TaskAVSRegistrarBase is
         AvsConfig memory _initialConfig
     ) internal onlyInitializing {
         __AVSRegistrar_init(_avs);
-        __Ownable_init();
-        _transferOwnership(_owner);
         _setAvsConfig(_initialConfig);
         __Allowlist_init(_owner);
     }
--- a/src/avs/task/TaskAVSRegistrarBase.sol
+++ b/src/avs/task/TaskAVSRegistrarBase.sol
@@ -1,18 +1,18 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.27;
 
-import {Initializable} from "@openzeppelin-upgrades/contracts/proxy/utils/Initializable.sol";
 import {IAllocationManager} from
     "eigenlayer-contracts/src/contracts/interfaces/IAllocationManager.sol";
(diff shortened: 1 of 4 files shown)


src/avs/task/TaskAVSRegistrarBase.sol
```diff
@@ -93,4 +97,22 @@ abstract contract TaskAVSRegistrarBase is
         avsConfig = config;
         emit AvsConfigSet(config.aggregatorOperatorSetId, config.executorOperatorSetIds);
     }
+
+    /// @notice Before registering operator, check if the operator is in the allowlist
+    /// @dev Reverts for:
+    ///      - OperatorNotInAllowlist: The operator is not in the allowlist
```
0xrajath: Small nit about the natspec following the same format as the other internal functions.

src/avs/task/TaskAVSRegistrarBase.sol
```diff
@@ -93,4 +97,22 @@ abstract contract TaskAVSRegistrarBase is
...
+    function _beforeRegisterOperator(
+        address operator,
+        uint32[] calldata operatorSetIds,
+        bytes calldata data
+    ) internal override {
+        super._beforeRegisterOperator(operator, operatorSetIds, data);
+
+        for (uint32 i; i < operatorSetIds.length; ++i) {
```
0xrajath:
```suggestion
        for (uint32 i = 0; i < operatorSetIds.length; i++) {
```

0xrajath: Small nit again to ensure that the for loops follow the same format.

nadir-akhtar: This is actually the syntax pulled straight from `AVSRegistrarWithAllowlist` -- can update here, but thought I'd call that out

src/avs/task/TaskAVSRegistrarBase.sol
```diff
@@ -22,7 +24,8 @@ abstract contract TaskAVSRegistrarBase is
     Initializable,
     OwnableUpgradeable,
     AVSRegistrarWithSocket,
-    TaskAVSRegistrarBaseStorage
+    TaskAVSRegistrarBaseStorage,
+    Allowlist
```
0xrajath: Small nit: Should we just move the `Allowlist` module before `TaskAVSRegistrarBaseStorage` in the inheritance chain?

nadir-akhtar: That's actually what I had before, but was getting a storage layout failure.

Since I don't believe we've deployed this to testnet, I don't think the storage layout change is actually a concern because of that, but thought if swapping it removes the failure, it would make sure we definitely don't have any conflicts because of the inheritance

0xrajath: Yeah the Storage layout is always going to fail since we'll have new storage slots

nadir-akhtar: Cool. My prior understanding was that it only fails if we clobber existing storage, but makes sense if any storage slot changes cause a failure. 

Probably a nice-to-have for the future since it's currently "required" to either make it a warning instead, or refactor it not to go red for "safe" storage updates (e.g. append only).

src/avs/task/TaskAVSRegistrarBase.sol
```diff
@@ -53,6 +56,7 @@ abstract contract TaskAVSRegistrarBase is
         __Ownable_init();
         _transferOwnership(_owner);
         _setAvsConfig(_initialConfig);
+        __Allowlist_init(_owner);
```
0xrajath: Small nit. Should this come before 

```
__Ownable_init();
_transferOwnership(_owner);
```

since the allowlist module has its own ownable set up?

nadir-akhtar: Good point -- refactor removing duplicative inheritance has this resolved

0xrajath (approved):
lgtm

0xrajath:
Storage Layout failure is expected due to change in storage slots

PR #542, opened 2025-10-10
feat: nonsigner view for table calc

eigenmikem:
**Motivation:**

Getting the index of an operator in an operator table and constructing nonsigner witnesses for multichain currently require unnecessarily complicated offchain code.

**Modifications:**

Two new views for default table calculators, helper functions to deduplicate code used by existing functions and new views

**Result:**

Better UX for multichain

### DIFF
--- a/src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
+++ b/src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
@@ -7,6 +7,8 @@ import {IOperatorTableCalculator} from
 import {IKeyRegistrar} from "eigenlayer-contracts/src/contracts/interfaces/IKeyRegistrar.sol";
 import {Merkle} from "eigenlayer-contracts/src/contracts/libraries/Merkle.sol";
 import {BN254} from "eigenlayer-contracts/src/contracts/libraries/BN254.sol";
+import {IBN254CertificateVerifierTypes} from
+    "eigenlayer-contracts/src/contracts/interfaces/IBN254CertificateVerifier.sol";
 import {LeafCalculatorMixin} from
     "eigenlayer-contracts/src/contracts/mixins/LeafCalculatorMixin.sol";
 import {IBN254TableCalculator} from "../../interfaces/IBN254TableCalculator.sol";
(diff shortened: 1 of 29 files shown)


src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
```diff
@@ -98,6 +100,140 @@ abstract contract BN254TableCalculatorBase is IBN254TableCalculator, LeafCalcula
...
+
+        // Early return if there are no weights/operators
+        if (weights.length == 0) {
+            return (false, 0);
+        }
+
+        uint32 includedIndex = 0;
+        for (uint256 i = 0; i < operators.length; i++) {
+            // Skip unregistered operators, as they are not part of the operator table
+            if (!keyRegistrar.isRegistered(operatorSet, operators[i])) {
```
ypatil12: We do this pattern of getting all the operators and then checking if they're registered across the contract. I'm wondering if we can create a helper function to reduce the duplication of code and make it so another contributor wouldn't forget in the future

src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
```diff
@@ -98,6 +100,140 @@ abstract contract BN254TableCalculatorBase is IBN254TableCalculator, LeafCalcula
...
+        BN254OperatorInfo[] memory includedOperatorInfos = new BN254OperatorInfo[](operators.length);
+        address[] memory includedOperators = new address[](operators.length);
+        uint256 includedCount = 0;
+
+        for (uint256 i = 0; i < operators.length; i++) {
+            if (!keyRegistrar.isRegistered(operatorSet, operators[i])) {
+                continue;
+            }
+
+            (BN254.G1Point memory g1Point,) = keyRegistrar.getBN254Key(operatorSet, operators[i]);
```
ypatil12: First part of this function can use `getOperatorInfos` ?

ypatil12: Also shares a lot of logic with `_calculateOperatorTable`. Can we make a helper function so code is deduped

ypatil12 (commented):
Can you add a PR description too

src/middlewareV2/tableCalculator/BN254TableCalculatorBase.sol
```diff
@@ -98,6 +100,86 @@ abstract contract BN254TableCalculatorBase is IBN254TableCalculator, LeafCalcula
...
+            );
+        }
+
+        // Prepare outputs with max length then resize
+        nonSignerWitnesses = new IBN254CertificateVerifierTypes.BN254OperatorInfoWitness[](
+            registeredOperators.length
+        );
+        nonSignerApk = BN254.G1Point(0, 0);
+
+        uint256 out = 0;
```
ypatil12: Nit: better variable name would be more descriptive here

ypatil12: Something like `nonSignerCount`

ypatil12 (approved):
Minor comment, but LGTM!

PR #544, opened 2025-11-03
feat: update registry coordinator for new `createOperatorSets`

ypatil12:
**Motivation:**

As part of https://github.com/Layr-Labs/eigenlayer-contracts/pull/1645, we are adding a new `createOperatorSets` function that passes in the slasher address upon creation of an operatorSet.  Although the *old* `createOperatorSets` function is not being deprecated, we are updating the `RegistryCoordinator` in order to be **only compatible the new function**. 

**Modifications:**

- Update `createSlashableStakeQuorum` to take in a `slasher` address
- Update all tests
- Update `solc` version in `foundry.toml` to `0.8.29`, matching Core

**Result:**

Compatible with updated `AllocationManager` interface

PR #548, opened 2025-12-01
feat: update registry coordinator for new createOperatorSets 

ypatil12:
**Motivation:**

As part of https://github.com/Layr-Labs/eigenlayer-contracts/pull/1645, we are adding a new `createOperatorSets` function that passes in the slasher address upon creation of an operatorSet.  Although the *old* `createOperatorSets` function is not being deprecated, we are updating the `RegistryCoordinator` in order to be **only compatible the new function**. 

**Modifications:**

- Update `createSlashableStakeQuorum` to take in a `slasher` address
- Update all tests
- Update `solc` version in `foundry.toml` to `0.8.29`, matching Core

**Result:**

Compatible with updated `AllocationManager` interface

### DIFF
--- a/.github/workflows/storage-report.yml
+++ b/.github/workflows/storage-report.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: v1.3.5
+          version: nightly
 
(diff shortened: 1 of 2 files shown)


ypatil12:
Prev PR was erroneously removed: https://github.com/Layr-Labs/eigenlayer-middleware/pull/544

